PR #220, opened 2016-09-02
Merge upstream changes from Nylas

bengotow:


mjs:
Thanks! This breaks the Gmail label tests, but I'll sort that out. Those tests could do with some work anyway.

PR #221, opened 2016-09-03
When raising a connection abort, try to include relevant ALERT sent by IMAP server

bengotow:


bengotow:
(Accidentally included other commits)

PR #222, opened 2016-09-08
On IMAP4 abort, include any prior ALERT message

bengotow:
We're seeing a large number of socket disconnections for a small subset of accounts on the sync engine while we're trying to perform "syncback" of changes. I think that we're probably receiving an unsolicited ALERT message from the server before the socket is closed. The goal of this patch is to capture that ALERT message so that we can see if there's more information we could use to debug the issue. (Hopefully this is a "too many commands too quickly", "this account has been suspended", etc., sort of response.)

It'd also be interesting to implement more generic support for ALERT messages—what do you think?

### DIFF
--- a/imapclient/imapclient.py
+++ b/imapclient/imapclient.py
@@ -1183,13 +1183,18 @@ class IMAPClient(object):
         uid = pop_with_default(kwargs, 'uid', False)
         assert not kwargs, "unexpected keyword args: " + ', '.join(kwargs)
 
-        if uid and self.use_uid:
-            if PY3:
-                command = to_unicode(command)  # imaplib must die
-            typ, data = self._imap.uid(command, *args)
-        else:
-            meth = getattr(self._imap, to_unicode(command))
-            typ, data = meth(*args)
+        try:
+            if uid and self.use_uid:
+                if PY3:
+                    command = to_unicode(command)  # imaplib must die
+                typ, data = self._imap.uid(command, *args)
+            else:
+                meth = getattr(self._imap, to_unicode(command))
+                typ, data = meth(*args)
+        except imaplib.IMAP4.abort as err:
+            alert = self._imap.untagged_responses.pop('ALERT', None)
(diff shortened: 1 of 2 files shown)


imapclient/imapclient.py
```diff
@@ -1183,13 +1183,18 @@ def _command_and_check(self, command, *args, **kwargs):
...
+            if uid and self.use_uid:
+                if PY3:
+                    command = to_unicode(command)  # imaplib must die
+                typ, data = self._imap.uid(command, *args)
+            else:
+                meth = getattr(self._imap, to_unicode(command))
+                typ, data = meth(*args)
+        except imaplib.IMAP4.abort as err:
+            alert = self._imap.untagged_responses.pop('ALERT', None)
+            raise imaplib.IMAP4.abort(err.message + (". Alert: %r" % alert))
```
mjs: Can you please change this slightly so the "Alert: ..." part only gets added on if there's actually an ALERT?

mjs:
Hi Ben - thanks for this.

Have you actually seen ALERTs in `untagged_responses`? (is this change already running in Nylas' fork of IMAPClient?) According to  https://tools.ietf.org/html/rfc3501#section-7.1 ALERTs are supposed to come back in the same line that has the `OK`, `NO` or `BAD` response from the server, not as unsolicited responses.

For example:
```
C: x select "foo"
S: x NO [ALERT] rate limiting applied
```

It's possible that IMAPClient / imaplib could be hiding these. If so, I'd like to get that fixed. It would be useful to find out if the syncback service is seeing a NO or BAD response to the last command before the disconnect.

That said, if you are seeing ALERTs in `untagged_responses` then I'm more than happy to merge this.

bengotow:
Ahh you're right—I was looking to duplicate the ALERT implementation in the NodeJS IMAP library (https://github.com/mscdex/node-imap/blob/master/lib/Connection.js#L1296), but it looks like they parse alerts within the OK response handler. I'll make some changes to this!

mjs:
This appears to be rotting so I'll close it for now. Feel free to re-open if there's updates.

PR #223, opened 2016-09-08
Tests for bad data in folder_status, more robust handling

bengotow:
In the last day we’ve seen this ValueError 249,915 times. It looks like the response can be more than two parts if A) the mailbox name contains spaces or B) the server decides to send us other data (or IMAPClient rolls it into the response buffer, not clear yet.) Either way, we should be able to handle these scenarios without crashing.

Relevant internal logs: https://nylas.loggly.com/search#terms=%22folder_status%20ValueError%22&from=2016-09-02T01%3A57%3A05.585Z&until=2016-09-03T01%3A57%3A05.585Z&source_group=

### DIFF
--- a/imapclient/imapclient.py
+++ b/imapclient/imapclient.py
@@ -635,10 +635,11 @@ class IMAPClient(object):
             what = normalise_text_list(what)
         what_ = '(%s)' % (' '.join(what))
 
-        data = self._command_and_check('status', self._normalise_folder(folder), what_)
+        fname = self._normalise_folder(folder)
+        data = self._command_and_check('status', fname, what_)
         response = parse_response(data)
         try:
-            _, status_items = response
+            status_items = response[-1]
         except ValueError:
             raise Exception('folder_status ValueError: Input ' + str(response))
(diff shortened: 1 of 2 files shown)


mjs:
Thanks for this. This looks good but I'd like to see an example of the data coming back from the server and I don't have access to Nylas' loggly. Could you include a example response here? Alternatively, which  email provider is triggering this?

bengotow:
Hey! Thanks for the feedback—the two samples I used in the new tests are values we were seeing in the logs. Unfortunately we didn't log the raw IMAP data, just the parsed data logged by https://github.com/mjs/imapclient/commit/93930936143b97d3ebe2200ea9fd0dee5db4ec08. (That commit was essentially step one to identifying a fix for this!) I can get you a bigger block of those log messages from loggly if you want.

mjs:
All good - merged!

mjs:
Note the followup tweaks to this in f994af6680d18442b54d8a7f4059f6c81d053413 and 77047bafd9a82f3bb1faa5e909d776b7a2e1b432

PR #224, opened 2016-12-12
Python 3.4 adjustments

frispete:
Hi Menno,

here's an attempt to adjust IMAPClient to the demands of Python 3.4. 
I tried to be careful with respect of backwards compatibility, without sacrifying the new achievements of Python 2.7.9 and 3.4 in terms of ssl. 

In that course, I decided to depend on the importability of backports.ssl. I expect, that current Python installations do not install this package, and adjusted setup.py accordingly.

While at it, I fixed the test fallout of 3.4, and removed the dependency on mock, if unittest.mock is available. 

setup.py test succeeds with 2.7.12 and 3.4.5 now.

This is just the initial PR, expect more to come this way, as I've started a project with a strong focus on IMAP last week, and decided to rely on IMAPClient, rather than imaplib.

### DIFF
--- a/imapclient/__init__.py
+++ b/imapclient/__init__.py
@@ -7,10 +7,17 @@
 
 from __future__ import unicode_literals
 
+from .version import author as __author__, version as __version__, version_info
 from .imapclient import *
 from .response_parser import *
-from .tls import *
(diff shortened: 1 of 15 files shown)


frispete:
Forget about this patch, will try again...

PR #225, opened 2016-12-13
Wire up Python 3.4/2.7.9 ssl module

frispete:
Hi Menno,

here's a new attempt to wire up IMAPClient to the ssl module of Python 3.4, 2.7.9.
I tried to be careful with respect of backwards compatibility, without sacrifying the new achievements of Python 2.7.9 and 3.4 in terms of ssl.

In order to provide a minimum invasive patch, I just changed tls.py to use the preferred ssl module, and adjusted setup.py accordingly.

While at it, I fixed some test fallout of 3.4, removed the dependency on mock, if unittest.mock is available, and adjusted some hg-sities to git.

setup.py test succeeds with 2.7.12 and 3.4.5 now.

This is just the initial PR, expect more to come this way, as I've started a project with a strong focus on IMAP last week, and decided to rely on IMAPClient, rather than imaplib.

I would be glad, if you consider to apply this PR.


mjs:
Thanks for this! I'll take a closer look next week once I'm off work.

I've actually been wondering about dropping the use of backports.ssl completely since it's been causing more trouble than it's worth (perhaps warning at import time if people are on pre 2.7.9/3.4 versions of Python). This PR is still useful in that case regardless.

frispete:
My pleasure, Menno.

After applying this, you won't get much in touch with the old cruft, given IMAPClient users do not live behind the moon. And if they do, they pay (or better: paid already) the price with the dependency to `backports.ssl` and `pyOpenSSL` - their choice. A deprecation warning would be a good amendment. I can provide one, once this PR is in.

I've tried several approaches to tackle this issue, and be quite happy with this one. The compatibility cruft is down to a minimum (2 lines for the new way, 1 line for the old way), and for those of us living in the present, it just takes the "good" road. All is well...

I've also thought about throwing out the cruft, but refrained from doing it, mostly because I don't wanted to upset **you** *too* much with the first PR, and I'm not a friend of *too* radical changes either ;)...

The changeset of `tls.py` looks scary, but that's just because I've indented the old part, while `IMAP4_TLS` is shared now. 

I'm hitting `IMAPClient` hard in a local setting, and it's doing well so far. What's missing compared to imaplib is the CRAM MD5 login scheme, but that's not *that* easy to implement (at least not in 15 minutes, that I spend on this today..). 

Cheers,
Pete

imapclient/imapclient.py
```diff
@@ -312,7 +312,7 @@ def capabilities(self):
         # If the server returned an untagged CAPABILITY response
         # (during authentication), cache it and return that.
         untagged = _dict_bytes_normaliser(self._imap.untagged_responses)
-        response = untagged.pop(b'CAPABILITY', None)
+        response = untagged.pop('CAPABILITY', None)
```
frispete: It was an attempt to fix a BytesWarning exception, that you harvest from a IMAPClient.capabilities() call with Python 3, running with -bb.

Now, that I look, is seems strange, because it was going though the _dict_bytes_normailizer before. Here's the traceback:
```
  File "imap4.py", line 62, in backup
    self.imap.capabilities()))
  File "imapclient/imapclient.py", line 315, in capabilities
    response = untagged.pop(b'CAPABILITY', None)
  File "imapclient/imapclient.py", line 1439, in pop
    return self._d.pop(k)
BytesWarning: Comparison between bytes and string
```
Okay, not that surprising, given, the keys yielded from _gen_keys may be both, strings and bytes. These are the times, where I regret to have been moved to python3..

imapclient/imapclient.py
```diff
@@ -1271,6 +1271,13 @@ def _normalise_labels(self, labels):
             labels = (labels,)
         return [_quote(l) for l in labels]
 
+    @property
+    def welcome(self):
```
frispete: Will fix

imapclient/test/test_datetime_util.py
```diff
@@ -6,7 +6,10 @@
 
 from datetime import datetime, date
 
-from mock import patch
+try:
+    from unittest.mock import patch
+except ImportError:
+    from mock import patch
```
frispete: Will fix accordingly

imapclient/tls.py
```diff
@@ -17,148 +18,181 @@
...
-    if V(installed_pyopenssl_version) < V(min_pyopenssl_version):
-       raise ImportError(
-           "pyOpenSSL version (%s) is too old. Need at least %s.\n"
-           "See http://imapclient.rtfd.io/#old-pyopenssl-versions"
-           % (installed_pyopenssl_version, min_pyopenssl_version))
+    import ssl
 
-if os.environ.get("READTHEDOCS") != "True":
-    check_pyopenssl_version()
+    create_default_context = ssl.create_default_context
```
frispete: Okay, sorry

frispete:
I hope, you don't mind the patches on top. 

BTW, try running the test suite this way: `python3 -bb setup.py test`

The Python3 induced bytes vs. str issue is awaiting some work still.

frispete:
Related to dropping compatibility: yes, that would allow simplifying some IMAPClient areas significantly…

If I have your blessing, I will go for it (rather soonish).

The other battlefield is the unholy bytes vs. str issue, which is biting me **all** the time.

This one needs a clear policy and I vote for eliminating bytes at the earliest possible moment (when they arrive), and convert back to it at the latest possible moment (when they leave). Everything in between should be strings. Exception from this rule are (RFC822) blobs, and possibly RFC822.HEADER (because MTA/MUA encoding glitches..). Apart from the (quirky-utf-7) Mailbox name encoding, it should all be nicely 'ascii' decodable.

mjs:
Merged! Thank you for this. 

I've created a `1.x` branch so that I can get the currently unreleased work out under 1.1.0 with the TLS changes. master is now `2.0-alpha`.

I've noticed one problem with the version checks in setup.py which means the tests don't work on Python 3.3 but I'll fix that now. 

I completely agree that eliminating bytes at the boundaries is the right approach. Unfortunately imaplib is quite consistent about this both between methods and between Python versions. This contributes to some of the weirdness. That said, I'm sure IMAPClient can do better. I welcome PRs or bug reports for where IMAPClient's behaviour is suboptimal this regard.

Thanks for the `python3 -bb` tip. I've just created https://github.com/mjs/imapclient/issues/226 to cover this.

Let me do some more thinking about dropping backports.ssl completely. I have some notes and a half complete blog article about TLS support across Python versions which I need to review.

mjs:
Minor fixes relating to this PR:

* https://github.com/mjs/imapclient/commit/8133a9ccd509a3f9e712c5c9cbfe0fa6bbb8e46e
* https://github.com/mjs/imapclient/commit/f17da6799bd47493f4263417f958e3d7e8cf1ead

frispete:
Hi Menno,

thank you for merging and the fixes, duh, my bad!

We're not in hurry. I will continue to send PRs when issues arise.

PR #229, opened 2017-01-06
provide default ssl_context, if not given explicitly

frispete:
Hi Menno,

here's a minimal invasive fix for a problem, that I stumbled over...

Problem is that datetime_to_INTERNALDATE relies on datetime formatting. datetime.strftime uses localized month abbreviations for %b, that fails in ugly ways for non en locales when calling append() with msg_time...

### DIFF
--- a/imapclient/datetime_util.py
+++ b/imapclient/datetime_util.py
@@ -50,7 +50,8 @@ def datetime_to_INTERNALDATE(dt):
     """
     if not dt.tzinfo:
         dt = dt.replace(tzinfo=FixedOffset.for_system())
-    return dt.strftime("%d-%b-%Y %H:%M:%S %z")
+    fmt = '%d-' + _SHORT_MONTHS[dt.month] + '-%Y %H:%M:%S %z'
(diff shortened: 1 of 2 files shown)


frispete:
the second patch set is an attempt to make starttls work (but it still fails during login..)

mjs:
Good catch! Thanks for this.

Can you please remove the second patch from this PR? I'd like to avoid mixing topics in one PR where possible.

mlorant:
@frispete Could you do that? If not, I might cherry-pick your changes to open a new PR to make things move forward :smile:

frispete:
@mlorant I'm pretty stuffed with things ATM, so please go ahead..

mlorant:
Ok, I will take care of it later and mention your name in the commitmsg, no problem :-)

mlorant:
Initial patch copied into a new PR: https://github.com/mjs/imapclient/pull/245

mjs:
Closing this as these changes have all been merged as separate PRs now.

Issue #217, opened 2016-08-16
Revert to using Python's built-in SSL/TLS support


mjs:
Originally reported by: **Menno Smits (Bitbucket: [mjs0](https://bitbucket.org/mjs0))**

---

Using backports.ssl, cryptography.io and pyOpenSSL for TLS support is proving to be more trouble than it's worth. Python's TLS support has been greatly improved since Python 2.7.9 and 3.4 and most people have access to these versions these days.

On older platforms fake it as best as possible, like urllib3 does:
https://github.com/kennethreitz/requests/blob/05d90b9379f57ee5f5d0beb268c495793fb5b2d7/requests/packages/urllib3/util/ssl_.py

Consider emitting warnings when on older versions.

---
- Bitbucket: https://bitbucket.org/mjs0/imapclient/issue/220

pilona:
Would this be a sufficient quick fix in the meantime?

    --- old.py	2016-09-27 12:54:33.256795485 -0400
    +++ tls.py	2016-09-27 12:56:21.319793550 -0400
    @@ -36,7 +36,11 @@
         check_pyopenssl_version()
     
     try:
    -    from backports import ssl
    +    if (version_info.major == 3 and tuple(version_info) < (3, 4) or
    +        version_info.major == 2 and tuple(version_info) < (2,7,9)):
    +        from backports import ssl
    +    else:
    +        import ssl
     except ImportError:
         raise ImportError("backports.ssl is not installed")
     
This would certainly unbreak TLS SNI support!

mjs:
#225 does most of the hard work to make this happen, using backports.ssl for older Python versions and the built-in ssl module for 2.7.9+ and 3.4+.

I'm still wondering about dropping backports.ssl, pyOpenSSL etc completely and just warning the user when they're on a Python version with crappy ssl support though.

boehlke:
New versions of imaplib provide a starttls command which works fine for me. Why not just use it if available (see my PR)

NicolasLM:
I would encourage you to drop `backports.ssl` altogether. A few reasons:

* It is unmaintained, which is a big red flag for a library that is about security
* It is installed by default with the wheel of `imapclient`, even on Python 3.6.1
* It makes debugging SSL issues very hard (backports.ssl vs PyOpenSSL vs  SSL in the standard library)
* The industry largely adopted Python 3.5 which has a working SSL implementation by default. Supporting older versions is not as important as it used to be.
[...]

PR #240, opened 2017-05-11
Use imaplib's starttls if available

boehlke:
This is an approach to #217

### DIFF
--- a/imapclient/imapclient.py
+++ b/imapclient/imapclient.py
@@ -192,17 +192,20 @@ class IMAPClient(object):
         Raises :py:exc:`AbortError` if the server does not support STARTTLS
         or an SSL connection is already established.
         """
-        if self.ssl or self._starttls_done:
-            raise self.AbortError('TLS session already established')
+        try:
+            return self._imap.starttls(ssl_context=ssl_context)
+        except AttributeError:
+            if self.ssl or self._starttls_done:
+                raise self.AbortError('TLS session already established')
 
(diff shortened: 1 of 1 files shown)


imapclient/imapclient.py
```diff
@@ -192,17 +192,20 @@ def starttls(self, ssl_context=None):
         Raises :py:exc:`AbortError` if the server does not support STARTTLS
         or an SSL connection is already established.
         """
-        if self.ssl or self._starttls_done:
-            raise self.AbortError('TLS session already established')
+        try:
+            return self._imap.starttls(ssl_context=ssl_context)
+        except AttributeError:
+            if self.ssl or self._starttls_done:
```
mjs: It's a little clearer IMO if the except clause was just a `pass` with the `if self.ssl ...` bit at the function level.

NicolasLM:
Since we will probably have to get rid of `imaplib` at some point, I am not sure that we should rely on another of its features.

mjs:
On reflection, I think @NicolasLM is right. We probably don't need this. Using Python's built in SSL support again has already been taken care of in a more general way. Do you agree @boehlke ?

mjs:
Closing pending further feedback. Feel free to re-open if you think it's necessary.

boehlke:
Hi, thanks for asking. Isn't this still an issue when using older Python versions? Please consider merging #314 ... someone might need the change. I added the change you proposed in my last commit.

Issue #218, opened 2016-08-17
Remove OAUTH(1) support


mjs:
Originally reported by: **Menno Smits (Bitbucket: [mjs0](https://bitbucket.org/mjs0))**

---

---
- Bitbucket: https://bitbucket.org/mjs0/imapclient/issue/221

PR #244, opened 2017-06-26
Fix #242: handle address without mailbox name or host in Address namedtuple

### LINKED ISSUE #242: Avoid TypeError when using `str` on Address tuple
Some emails have no mailbox or host (e.g. `undisclosed-recipients` case), so when parsing the ENVELOPE of the message using imapclient, we can get something like this:

```
In [8]: from imapclient.response_types import *
In [9]: addr = Address('Anonymous', None, None, 'undisclosed-recipients')
In [10]: str(addr)
---------------------------------------------------------------------------
.../lib/python3.5/site-packages/imapclient/response_types.py in __str__(self)
     57         return formataddr((
     58             to_unicode(self.name),
---> 59             to_unicode(self.mailbox) + '@' + to_unicode(self.host)))
     60 
     61 

TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'
```

I think the `__str__` method should handle this and just returning `self.mailbox` or `self.host` if the other part is missing. I could write the PR but I prefer to have thoughs about this before.

mlorant:
See the issue #242, not much to add :-)

### DIFF
--- a/imapclient/response_types.py
+++ b/imapclient/response_types.py
@@ -80,9 +80,12 @@ class Address(namedtuple("Address", "name route mailbox host")):
     """
 
     def __str__(self):
-        return formataddr((
-            to_unicode(self.name),
-            to_unicode(self.mailbox) + '@' + to_unicode(self.host)))
+        if self.mailbox and self.host:
+            address = to_unicode(self.mailbox) + '@' + to_unicode(self.host)
+        else:
+            address = to_unicode(self.mailbox or self.host)
+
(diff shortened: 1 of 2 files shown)


mjs:
OAUTH1 support was removed in #218 . Would you mind rebasing so this PR just has the `Address` fix?

mlorant:
@mjs Sorry for the delay, summertime!

Done, feel free to review.

mjs:
Merged into the `master` and `1.x` branches

Issue #230, opened 2017-01-06
starttls() doesn't work


frispete:
```
--> import imapclient
--> imap = imapclient.IMAPClient('192.168.x.y', 143)
--> imap.starttls()
--> imap.login('user', 'password')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "imapclient/imapclient.py", line 215, in login
    unpack=True,
  File "imapclient/imapclient.py", line 1193, in _command_and_check
    typ, data = meth(*args)
  File "/usr/lib64/python3.4/imaplib.py", line 538, in login
    typ, dat = self._simple_command('LOGIN', user, self._quote(password))
  File "/usr/lib64/python3.4/imaplib.py", line 1136, in _simple_command
    return self._command_complete(name, self._command(name, *args))
  File "/usr/lib64/python3.4/imaplib.py", line 959, in _command_complete
    typ, data = self._get_tagged_response(tag)
  File "/usr/lib64/python3.4/imaplib.py", line 1079, in _get_tagged_response
    self._get_response()
  File "/usr/lib64/python3.4/imaplib.py", line 987, in _get_response
    resp = self._get_line()
  File "/usr/lib64/python3.4/imaplib.py", line 1094, in _get_line
    if not line.endswith(b'\r\n'):
TypeError: endswith first arg must be str or a tuple of str, not bytes
```
I've tried to wrap my head around this problem, but lost myself in the maze of IMAPClient and imaplib, bytes vs. strings entanglements...

jcea:
Hitting this bug too. Quite critical, indeed. Any time schedule for publishing an updated release to PYPI?.

Thanks for your time and effort!

maparent:
Solved in #247

mjs:
@jcea: which version of IMAPClient are you hitting this with? My understanding was that this only happens with the unreleased `master` branch (which will become 2.x). Are you see this with an official release?

NicolasLM:
Closing as everything seems to work on master with Python 3:

```python
>>> import imapclient
>>> s = imapclient.IMAPClient("ssl0.ovh.net", 143, ssl=False)
>>> s.capabilities()
(b'IMAP4REV1', b'LITERAL+', b'SASL-IR', b'LOGIN-REFERRALS', b'ID', b'ENABLE', b'IDLE', b'STARTTLS', b'AUTH=PLAIN', b'AUTH=LOGIN')
>>> s.starttls()
b'Begin TLS negotiation now.'
[...]

PR #247, opened 2017-07-13
Get a binary file from the imap socket when doing TLS

maparent:
Solves #230


mjs:
Thank you very much for this. Would you mind removing the second commit from this PR (d2bc189a83a316145dc9f4cb23f2fa1e49f6504e) as this fix has already been merged separately.

maparent:
rebased on master, that commit is done. Cheers!

mjs:
Thank you!

PR #251, opened 2017-08-02
Officially support Python 3.6

NicolasLM:
We successfully run imapclient with Python 3.6 on production for a few months already.


mjs:
Great. Thank you!

PR #252, opened 2017-08-02
Use the standard logging module

NicolasLM:
Instead of a custom way of managing log lines, this commit
leverages the standard logging module. As IMAPClient is a library,
the only thing to do is to create a logger and log away.

The final user of IMAPClient is now in charge of formatting,
writing to files or discarding log lines.

### DIFF
--- a/doc/src/index.rst
+++ b/doc/src/index.rst
@@ -258,6 +258,23 @@ TLS Support
 .. automodule:: imapclient.tls
    :members:
 
+Logging
+~~~~~~~
+IMAPClient logs debug lines using the standard Python `logging module
+<https://docs.python.org/3/library/logging.html>`_. Its logger is
+``imapclient.*``.
+
+A simple way to display log messages is to setup logging::
+
+  import logging
+
+  logging.basicConfig(
+      format='%(asctime)s - %(levelname)s: %(message)s',
+      level=logging.DEBUG
+  )
+
+For advanced usage please refer to the Python documentation.
+
 Interactive Sessions
 --------------------
 When developing program using IMAPClient is it sometimes useful to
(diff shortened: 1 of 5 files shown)


doc/src/index.rst
```diff
@@ -258,6 +258,23 @@ TLS Support
...
+A simple way to display log messages is to setup logging::
+
+  import logging
+
+  logging.basicConfig(
+      format='%(asctime)s - %(levelname)s: %(message)s',
+      level=logging.DEBUG
+  )
+
+For advanced usages please refer to the Python documentation.
```
mjs: s/usages/usage/

imapclient/imapclient.py
```diff
@@ -28,6 +29,9 @@
     long = int  # long is just int in python3
 
 
+logger = getLogger(__name__)
+imaplib_logger = getLogger('imapclient.imaplib')
```
mjs: What about `__name__ + ".imaplib"`?

NicolasLM: Then the logger would be called `imapclient.imapclient.imaplib`. I initially named it only `imaplib` since it would be its name if imaplib was using logging.

What do you prefer?

NicolasLM:
Do you want to release this for `1.1.0` or `2.0.0`? Strictly speaking this is a breaking change, albeit a small one.

imapclient/imapclient.py
mjs: I didn't realise that "imapclient" would be doubled up. The way you have it is better.

doc/src/releases.rst
```diff
@@ -26,6 +26,8 @@ Changed
 -------
 - the `mock` package is no longer installed by default (just as a test
   dependency)
+- Logs are now handled by the Python logging module. `debug` and `log_file`
```
mjs: Let's release this in 2.x, not 1.x, given that it is a breaking change.

NicolasLM: Done

Issue #257, opened 2017-08-07
Drop support for Python 2.6 and Python 3.3


mlorant:
Python 2.6 is not officially supported anymore since October 2013. Python 3.3 official support [ends this September](https://www.python.org/dev/peps/pep-0398/#lifespan).

I know there is some code that checks Python version [there that shoud be kept](https://github.com/mjs/imapclient/blob/1503e95c8154fb21805e6b4eb8acfaae2ef07442/imapclient/tls.py#L22) (for 3.4 support) and some [here that could be throw away](https://github.com/mjs/imapclient/blob/77047bafd9a82f3bb1faa5e909d776b7a2e1b432/imapclient/imaplib_ssl_fix.py) (2.6 only fix).

I can make a PR for both but it should be merged in September only, so we don't anticipate the end of support of the 3.3 too early. Otherwise two distinct PR would also work :-)

NicolasLM:
I think both can be targeted at 2.0.

Since no code would be removed by dropping support for 3.3, IMAPClient will still be installable on 3.3, just "at your own risk".

mjs:
Thank you for the thorough patch. I've asked about this on the mailing list but I'm personally very happy for this to land.

I think it's fine to land in `master` today. 2.x is unlikely to be released before Python 3.3 is no longer supported.

mjs:
Done in #258.

PR #258, opened 2017-08-07
Drop support of Python 2.6 and 3.3

mlorant:
See issue #257


mlorant:
When working on my other PR about ConfigParser import, I've seen an artefact of Python < 2.6 support.
In `imapclient.config.py`: `json` module is provided since Python 2.6 so the `except ImportError` should not be necessary now. 

Let me fix this before merging.

PR #260, opened 2017-08-08
Use six instead of try/except for ConfigParser import

mlorant:
To reply to the deleted comment `# TODO: can't six do this for us?` in the codebase, yes we can.


mlorant:
(I have a broken gitconfig on this PC between my business email and my personal one, sorry about that if it mess up with commit author...)

mjs:
Great!

Issue #200, opened 2016-02-10
UNSELECT support


mjs:
Originally reported by: **Erik Quaeghebeur (Bitbucket: [equaeghe](https://bitbucket.org/equaeghe))**

---

See http://tools.ietf.org/html/rfc3691.

Interface ideas: by adding a new method, or by making select_folder(None) do this.

Fall-back for servers not supporting this (few) is, as per the RFC:

> [IMAP4] provides this functionality (via a SELECT command with a nonexistent mailbox name or reselecting the same mailbox with EXAMINE command)

This kind of functionality may be useful for clients to release resources for some reason or other.

---
- Bitbucket: https://bitbucket.org/mjs0/imapclient/issue/203

mjs:
Done!

PR #263, opened 2017-08-09
#200 - Add UNSELECT command

mlorant:
As said in the commitmsg, this is still a work in progress. Things missing:

- [x] More tests (especially one or two unit tests)
- [x] Add a `has_capability` check on the live test in case the server is not supporting it
- [x] Review the public API chosen (see below)
- [x] Documentation on `unselect_folder` or `select_folder(None)` (see below)

Does `unselect_folder()` is useful as a public method if we favor the `select_folder(None)`? I would prefer to keep a single _public_ way to use it. 
My two cents: I like the `(None)` one (it reminds me [Django convetion for select_related/prefetch_related](https://docs.djangoproject.com/en/1.11/ref/models/querysets/#select-related)) but `unselect_folder` as a separate method can be valid since it is a different IMAP command...

Issue reference: #200

### DIFF
--- a/imapclient/imapclient.py
+++ b/imapclient/imapclient.py
@@ -53,6 +53,10 @@ if 'STARTTLS' not in imaplib.Commands:
 if 'ID' not in imaplib.Commands:
     imaplib.Commands['ID'] = ('NONAUTH', 'AUTH', 'SELECTED')
 
+# ... and UNSELECT. RFC3691 does not specify the state but there is no
+# reason to use the command without AUTH state and a mailbox selected.
+if 'UNSELECT' not in imaplib.Commands:
+    imaplib.Commands['UNSELECT'] = ('AUTH', 'SELECTED')
 
 # System flags
 DELETED = br'\Deleted'
@@ -471,6 +475,19 @@ class IMAPClient(object):
         """
         self._command_and_check('select', self._normalise_folder(folder), readonly)
         return self._process_select_response(self._imap.untagged_responses)
+
+    def unselect_folder(self):
+        """Unselect the current folder and release associated resources.
+
+        Unlike ``close_folder``, the ``UNSELECT`` command does not expunge
+        the mailbox, keeping messages with \Deleted flag set for example.
+
+        Returns the UNSELECT response string returned by the server.
+        """
+        logger.debug('< UNSELECT')
+        # IMAP4 class has no `unselect` method so we can't use `_command_and_check` there
+        _typ, data = self._imap._simple_command("UNSELECT")
+        return data[0]
 
     def _process_select_response(self, resp):
         untagged = _dict_bytes_normaliser(resp)
--- a/imapclient/livetest.py
+++ b/imapclient/livetest.py
@@ -234,6 +234,22 @@ class TestGeneral(_TestBase):
         self.assertEqual(ns.other, ns[1])
         self.assertEqual(ns.shared, ns[2])
 
+    def test_unselect_folder(self):
+        if not self.client.has_capability('UNSELECT'):
+            return self.skipTest("Server doesn't support UNSELECT")
+
+        resp = self.client.select_folder(self.base_folder)
+        self.assertEqual(resp[b'EXISTS'], 0)
+        self.client.search(['ALL'])
+        self.client.unselect_folder()
+
+        # To ensure the folder has been selected, check we can't run .search()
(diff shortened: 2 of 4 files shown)


mlorant:
For information, I have chosen to keep `unselect_folder`: easier for documentation and keep the spirit of "1 IMAP command = 1 method"

NicolasLM:
Good choice, I also prefer having a separate command.

imapclient/livetest.py
```diff
@@ -234,6 +234,17 @@ def assertNoneOrTuple(val):
...
         self.assertEqual(ns.shared, ns[2])
 
+    def test_select_and_unselect(self):
+        if not self.client.has_capability('UNSELECT'):
+            return self.skipTest("Server doesn't support UNSELECT")
+
+        resp = self.client.select_folder(self.base_folder)
+        self.assertEqual(resp[b'EXISTS'], 0)
+        self.client.search(['ALL'])
+        resp = self.client.select_folder(None)
```
NicolasLM: Don't you mean `self.client.unselect_folder()`?

mlorant: Obviously, yes, You got me: I did not even re-run the test after removing the `None` option...

imapclient/imapclient.py
```diff
@@ -472,6 +476,22 @@ def select_folder(self, folder, readonly=False):
         self._command_and_check('select', self._normalise_folder(folder), readonly)
         return self._process_select_response(self._imap.untagged_responses)
 
+    def unselect_folder(self):
+        """Unselect the current folder and release ressources associated.
```
mjs: typo: ressources 

also: "release associated resources" is probably flows better (swap the words).

imapclient/imapclient.py
```diff
@@ -472,6 +476,22 @@ def select_folder(self, folder, readonly=False):
...
+
+        Unlike ``close_folder``, the ``UNSELECT`` command does not expunge
+        the mailbox, keeping messages with \Deleted flags for example.
```
mjs: "... keeping messages with the \Deleted flag set."

imapclient/imapclient.py
```diff
@@ -472,6 +476,22 @@ def select_folder(self, folder, readonly=False):
...
+
+        The return value is the server command response message
+        followed by a list of status responses, which should be empty.
```
mjs: When wouldn't the list of status responses be empty? Why return it if it's always empty?

imapclient/livetest.py
```diff
@@ -234,6 +234,17 @@ def assertNoneOrTuple(val):
...
+    def test_unselect_folder(self):
+        if not self.client.has_capability('UNSELECT'):
+            return self.skipTest("Server doesn't support UNSELECT")
+
+        resp = self.client.select_folder(self.base_folder)
+        self.assertEqual(resp[b'EXISTS'], 0)
+        self.client.search(['ALL'])
+        resp = self.client.unselect_folder()
+        with self.assertRaises(IMAPClient.Error):
+            self.client.search(['ALL'])
```
mjs: Thank you so much for taking the time to extend the live tests too.

imapclient/livetest.py
```diff
@@ -234,6 +234,17 @@ def assertNoneOrTuple(val):
...
         self.assertEqual(ns.shared, ns[2])
 
+    def test_unselect_folder(self):
+        if not self.client.has_capability('UNSELECT'):
+            return self.skipTest("Server doesn't support UNSELECT")
+
+        resp = self.client.select_folder(self.base_folder)
+        self.assertEqual(resp[b'EXISTS'], 0)
+        self.client.search(['ALL'])
+        resp = self.client.unselect_folder()
```
mjs: No need to assign to `resp` if you're not checking it.

mjs:
Would you mind also including an addition to the release notes?

imapclient/imapclient.py
mlorant: I am still new at reading RFC spec in such details... I took a look at other methods and indeed, a lot of them have the same behaviour (such as LOGIN/COPY/NOOP...), so I will do the same here: returning only the string response.

mlorant:
I added the release notes in the 1.1.0 release, since the #200 issue has the "minor" label, and it is fully backward compatible since it is almost just one method insertion.

mjs:
This looks ready to me. I'll merge and then get it into the 1.x branch too.

Issue #141, opened 2015-02-28
Use more precise exception classes


mjs:
Originally reported by: **Menno Smits (Bitbucket: [mjs0](https://bitbucket.org/mjs0))**

---

For example, instead of IMAPClient raising an exception like:

```
imaplib.error: select failed: "Mailbox doesn't exist: days"
```

It would be better if a specific "mailbox doesn't exist" error was used. The same applies to other kinds of errors.

Requested by Thomas Güttler.

---
- Bitbucket: https://bitbucket.org/mjs0/imapclient/issue/144

NicolasLM:
Closing as it is related to #267

PR #267, opened 2017-08-10
#141 - Use precise exceptions for each type of error

mlorant:
This is a quick draft but I want a first approval before continuing this way, since it will require a lot of code to adapt.

I have defined a new module `imapclient.exceptions` that defines all types of exceptions that IMAPClient might raise. This is very similar [to what requests/django/pytz are doing](https://github.com/requests/requests/blob/master/requests/exceptions.py). The base class is actually just `imap.IMAP4.error` so we can still rely on `except imap.IMAP4.error` if we don't care at all about the source of the exception (+ backward compatibility with 1.x code) 

The most-time consuming is now to define what exceptions we need (I wrote down a couple of them) and raise them in the appropriate methods. 
Most of the exceptions raised in IMAPClient are actually exceptions that bubble up from imaplib itself. So we might need to catch them and re-raise with our more dedicated exception to make debugging and error handling more efficient for IMAPClient users.

### DIFF
--- a/imapclient/exceptions.py
+++ b/imapclient/exceptions.py
@@ -1,0 +1,38 @@
+import imaplib
+
+# Base class allowing to catch any IMAPClient related exceptions
+# To ensure backward compatibility, we "rename" the imaplib general
+# exception class, so we can catch its exceptions without having to
+# deal with it in IMAPClient codebase
+IMAPClientException = imaplib.IMAP4.error
+IMAPClientAbortError = imaplib.IMAP4.abort
+IMAPClientReadOnlyError = imaplib.IMAP4.readonly
+
+
+class CapabilityError(IMAPClientException):
+    """ 
+    The command tried by the user needs a capability not installed 
+    on the IMAP server
+    """
+
+
+class LoginError(IMAPClientException):
+    """
+    A connection has been established with the server but an error 
+    occurred during the authentication.
+    """
+
+
+class IllegalStateException(IMAPClientException):
+    """
+    The command tried needs a different state to be executed. This
+    means the user is not logged in or the command needs a folder to
+    be selected.
+    """
+
+
+class InvalidCriteriaException(IMAPClientException):
+    """ 
+    A command using a search criteria failed, probably due to a syntax 
+    error in the criteria string.
+    """
--- a/imapclient/imapclient.py
+++ b/imapclient/imapclient.py
@@ -16,6 +16,7 @@ from logging import LoggerAdapter, getLogger
 
 from six import moves, iteritems, text_type, integer_types, PY3, binary_type, iterbytes
 
+from . import exceptions
 from . import imap4
 from . import response_lexer
 from . import tls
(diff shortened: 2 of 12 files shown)


imapclient/imapclient.py
```diff
@@ -109,8 +110,6 @@ class IMAPClient(object):
     calls if required.
 
     """
-
-    Error = imaplib.IMAP4.error
```
mlorant: Maybe this should stay for backward-compatibility after all? Maybe a bit different with `Error = exceptions.IMAPClientException` but there should be no difference for people catching this in old versions.

imapclient/imapclient.py
```diff
@@ -535,7 +535,7 @@ def idle(self):
         self._idle_tag = self._imap._command('IDLE')
         resp = self._imap._get_response()
         if resp is not None:
-            raise self.Error('Unexpected IDLE response: %s' % resp)
+            raise exceptions.IDLEResponseError('Unexpected IDLE response: %s' % resp)
```
NicolasLM: I'm not sure about this one. It seems that it's just a command that failed. Is it different from any other command failing for a random reason?

mjs: I agree. I'm not sure this needs a separate exception type.

imapclient/imapclient.py
mjs: I'm in 2 minds on this. I'm not fond of having the exceptions on the class as it's a little unconventional/surprising, but I'm also worried about breaking everyone's code.

Perhaps leave `Error` there but don't document it?

imapclient/exceptions.py
```diff
@@ -0,0 +1,35 @@
...
+class NoSelectedFolder(IMAPClientException):
+    """ A folder needs to be selected before executing the command """
+
+
+class ReadOnlyError(IMAPClientException):
+    """ Raised if a modifying operation was attempted on a read-only folder """
+
+
+class InvalidSearchCriteria(IMAPClientException):
+    """ The SEARCH command failed, probably due to wrong criteria """
```
mjs: Perhaps generalise this to `InvalidCriteriaError` so that the name work better for `sort()` and `thread()` criteria too?

imapclient/imapclient.py
mlorant: This would be a good compromise indeed.

The "Exceptions" section in the doc will need to be rewritten to describe each exception class created. I will sligthly change the documentation of `IMAPClient`'s `Error`, `AbortError` and `ReadOnlyError` subclasses, as they will be now considered as classes of `imapclient.exceptions` module.
The `IMAPClient` will still have references to thoses three base classes, for backward compatibility.

mjs: That sounds great.

NicolasLM:
Having that for 2.0.0 would be awesome! @mlorant is there anything blocking you? Time I guess ;)

Let me know if there is anything I can help you with.

mlorant:
@NicolasLM Yes, I had some busy days these last weeks. I'll try to finish this in the last few days, I will have time this week-end, there's not so much left.

mlorant:
I have introduced some specific exceptions, but I won't have the courage to make a full list :-) I guess it could be introduced in future versions. This PR is a base for future implementation as it defines base exceptions, some specific common ones and remove the public doc about the legacy exceptions defined in IMAPClient (while they are still there for backward compatibility).

Some exceptions might be hard to raised without analyzing imaplib underlaying exceptions, that are way too generic sometimes... it means it would need to get rid of imaplib someday (something already written in another issue IIRC)

Issue #255, opened 2017-08-06
Document tips for handling large mailboxes


mjs:
* Use ENVELOPE first
* Chunk message ids

Issue #119, opened 2013-06-05
Move almost all docs to Sphinx


mjs:
Originally reported by: **Menno Smits (Bitbucket: [mjs0](https://bitbucket.org/mjs0))**

---

Integrate README and HACKING content into Sphinx docs and make them stubs which refer to the real docs. This means there's only one place people need to look and less for me to maintain.

---
- Bitbucket: https://bitbucket.org/mjs0/imapclient/issue/122

freddrake:
Isn't this substantially done?  There are only a few things left in the README.

NicolasLM:
Right, closing. Thanks!

PR #269, opened 2017-08-15
Split the documentation into several pages

mlorant:
Today, the documentation is one long page containing a mix of intro, tips and API reference. It is quite fine today but some issues will lengthen this page, such as #119, #255 and some best practices like "Contributing guide".

In this PR, I have split the `index.rst` into several files, inspired by popular Python packages. I have add some examples of titles of new sections not yet written but possible in the user guide. This results in an index lighter but with some tables of content that guide the user. There's a lot of changes but it is mostly cut and paste section there and there. Except the example in index.rst and some title hierarchy, I have not modified the content itself.

This split is also the opportunity to introduce some quickstart section, that show working example of the library and some guide to how handle most common cases. Like [requests simplest example](https://github.com/requests/requests/blob/master/docs/index.rst#requests-http-for-humans), I tried to make an example with and without IMAPClient, but ironically, I got bored just parsing imaplib responses by hand... 😑 

Finally, I have changed the example given in index.rst to give something more appealing. Without any knowledge of the IMAP protocol, the `\Deleted` flag is little obscure (typical reaction: _why deleted messages appears in the inbox?!_), so something simpler that displays a list of message filtered is more relevant I think.

I have some questions:

- Most of other libraries have their releases notes in the root of the repo while we have to refer to the documentation here... I'm mixed about this choice.
- Does anyone have some subsections ideas that would be useful in the documentation? A thread could be started on the ML to have more feedback.


mjs:
This looks awesome. Thank you thank you!

I'll take a proper look when I have more time, but it looks like you've done a lot of things I had been considering and then some.

AUTHORS.rst
```diff
@@ -64,31 +65,30 @@ Aviv Salem
...
+- Benjamin Morrise
+- Brandon Rhodes
+- Thomi Richards
+- Ivan Zderadicka
+- Raffael Vogler
+
+
+Other Contributions
+~~~~~~~~~~~~~~~~~~~
 IMAPClient uses Michael's excellent 'Mock' package.
```
NicolasLM: I would remove this section, most projects use other libraries as dependencies without mentioning all of them. Keeping this list up to date is cumbersome as well.

mlorant: I tend to agree. I would also remove the "who did what" sentences in the Code section, which is hard to maintain and just keep the contributor list in alphabetic order, with active maintainers at the top of the list. 
I would also remove the Bug Reports section which is not really "authors" but more thankful users.

mjs: +1. Let's:
* remove the specific mention of Mock
* remove the specifics of "who did what"
* remove the bug reports section

mjs:
Regarding where to put the release notes ... I'd prefer them to be in the root of the repo too but I'd also like them as a section in the docs. Is it possible to include them from the root of the repo?

Regarding other documentation sections ... I'm happy for this to be extended later (beyond this PR). One thing I've always wanted to include is a list of all the IMAP RFCs that IMAPClient supports with links to the tools.ietf.org.

mjs:
I've just pulled this PR and built the docs to get a better feel for how it looks. I *love* the new layout. TYVM.

My only concern is whether it makes sense for the "Interactive Sessions" part to be included in "Quickstart". It doesn't seem like something that people would necessarily need to be concerned with when getting started.

"Quickstart" should probably have the "Installing IMAPClient" section (as now), the "Connect to a server and send our first commands" bit (still to be done - maybe this could start very simple and then gradually explain some more complex things - I'm happy to write this) and then links to the API reference and further examples.

AUTHORS.rst
```diff
@@ -0,0 +1,32 @@
...
+| Thomas Steinacher
+| Zac Witte
+
+Other Contributions
+~~~~~~~~~~~~~~~~~~~
+Nylas supports development of IMAPClient.
+http://nylas.com/
+
+The project was started while the author was working at NetBox Blue.
+http://netboxblue.com/
```
mjs: This section can go. Nylas no longer supports the project and NetBox Blue was a *very* long time ago.

doc/src/contributing.rst
```diff
@@ -79,10 +78,16 @@ IMAPClient (provided these versions of Python are installed!).
...
 should typically be used as the base class for any tests - it provides
 a mock IMAPClient instance at `self.client`. See the tests in
 imapclient/tests/test_imapclient.py for examples of how to write unit
 tests using this approach.
+
+
+Authors and acknowledgment
+--------------------------
+
+.. include:: ../../AUTHORS.rst
```
mjs: I wonder if this section should be it's own top level section in index.rst instead of being here.

CONTRIBUTING.rst
```diff
@@ -0,0 +1 @@
+.. include:: doc/src/contributing.rst
```
mjs: Is this actually useful? I'd prefer that CONTRIBUTING.rst didn't exist here.

mlorant: This include does not work anyway. Will remove it.

mlorant: Does not work on Github * (on mobile, no edit option, d'un)

doc/src/contributing.rst
mlorant: Given that (AUTHORS + THANKS) files have been merged and reduced from 100 lines to 22, I can now include it in `index.rst`, indeed.

mlorant:
Just made another rebase/small fixes to take into account your last review and your commit about _freshfoo.com_ -> _menno.io_. 

As you said, we can merge it and fill the "TODO" parts later, so other PR can rely on the new structure.

PR #270, opened 2017-08-18
Prevent IMAP passwords from going to the logging system

NicolasLM:
Even if the final user did not enable DEBUG logging for its
application, some logs records can end up in tools like Sentry
via breadcrumbs.

This commit makes sure that passwords are never sent to the
logging system by verifying and redacting the log message before
passing it to the logger.

In the future, once we get rid of imaplib, the code replacing it
could avoid logging raw data containing sensitive information in
the first place.

### DIFF
--- a/imapclient/imapclient.py
+++ b/imapclient/imapclient.py
@@ -12,7 +12,7 @@ import sys
 import re
 from datetime import datetime, date
 from operator import itemgetter
-from logging import getLogger
+from logging import LoggerAdapter, getLogger
 
 from six import moves, iteritems, text_type, integer_types, PY3, binary_type, iterbytes
 
@@ -30,7 +30,6 @@ if PY3:
 
 
 logger = getLogger(__name__)
-imaplib_logger = getLogger('imapclient.imaplib')
 
 __all__ = ['IMAPClient', 'DELETED', 'SEEN', 'ANSWERED', 'FLAGGED', 'DRAFT', 'RECENT']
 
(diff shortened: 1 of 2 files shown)


mlorant:
Awesome, I did something equivalent because of Sentry too. However, could you also support `oauth2_login` and strip the user and access_token in that case? Thanks!

NicolasLM:
I added the same thing for the `AUTHENTICATE` command.

imapclient/test/test_imapclient.py
```diff
@@ -365,6 +366,26 @@ def test_IMAP_is_patched(self):
...
+            "> b'ICHH1 LOGIN **REDACTED**",
+            (),
+            extra={}
+        )
+        py2_called = logger_mock.info.called_once_with(
+            "> b'ICHH1 LOGIN **REDACTED**",
+            (),
+            extra={}
+        )
+        assert py2_called or py3_called, "Log message not redacted properly"
```
mjs: Why not use `six.PY3` to determine the Python version and choose the assert?

NicolasLM: Done

imapclient/test/test_imapclient.py
```diff
@@ -365,6 +366,27 @@ def test_IMAP_is_patched(self):
...
+        if six.PY3:
+            logger_mock._log.assert_called_once_with(
+                logging.INFO,
+                "> b'ICHH1 LOGIN **REDACTED**",
+                (),
+                extra={}
+            )
+        else:
+            logger_mock.info.assert_called_once_with(
+                "> b'ICHH1 LOGIN **REDACTED**",
```
mlorant: no `b'` prefix in py2, but I think the CI error would have help you too

NicolasLM: Fixed, thanks!

mjs:
I'm happy to merge this but there's a conflict.

NicolasLM:
Rebased

mjs:
Thanks again.

PR #271, opened 2017-08-19
Add ENABLE support

### LINKED ISSUE #136: Support ENABLE extension

---

https://tools.ietf.org/html/rfc5161

Use this to turn on CONDSTORE in livetests (better than current hack for Dovecot)

---
- Bitbucket: https://bitbucket.org/mjs0/imapclient/issue/139

mjs:
... and use it in the integration tests. 

Fixes #136.

### DIFF
--- a/imapclient/imapclient.py
+++ b/imapclient/imapclient.py
@@ -298,6 +298,8 @@ class IMAPClient(object):
             uid=False,
             response_name='ENABLED',
             unpack=True)
+        if not resp:
+            return []
         return resp.split()
 
     def id_(self, parameters=None):
--- a/imapclient/imapclient.py
+++ b/imapclient/imapclient.py
@@ -39,11 +39,11 @@ __all__ = ['IMAPClient', 'DELETED', 'SEEN', 'ANSWERED', 'FLAGGED', 'DRAFT', 'REC
 
 # We also offer the gmail-specific XLIST command...
 if 'XLIST' not in imaplib.Commands:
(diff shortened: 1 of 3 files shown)


imapclient/imapclient.py
```diff
@@ -268,6 +273,33 @@ def shutdown(self):
...
+        if self._imap.state != 'AUTH':
+            raise self.Error("ENABLE command illegal in state %s" % self._imap.state)
+
+        resp = self._raw_command_untagged(
+            b'ENABLE',
+            [to_bytes(c) for c in capabilities],
+            uid=False,
+            response_name='ENABLED',
+            unpack=True)
+        return resp.split()
```
NicolasLM: If the `ENABLE` command does not contain any capability that the server can actually enable, `None` is returned:

```python
>>> c.enable("FOO", "ID")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/nicolas/dev/imapclient/imapclient/imapclient.py", line 301, in enable
    return resp.split()
AttributeError: 'NoneType' object has no attribute 'split'
```

mjs: Weird - in my testing, an empty string was returned when the server wasn't able to enable anything. Must be a imaplib or server response difference between our tests.

I will modify the code to account for this.

mjs: I was able to reproduce what you saw using Fastmail. Fastmail doesn't send an ENABLED response when nothing was enabled, whereas Gmail sends an empty ENABLED response.

PR #274, opened 2017-08-21
Move all non-library code outside of the imapclient package

### LINKED ISSUE #266: Move tests out of the imapclient package
Most projects have their tests out of the main package, in a separate `test` package. See for instance requests, flask and celery.

This prevents tests from being included in the distribution (even if there are other ways to achieve this) and is more consistent with the practices of the Python ecosystem.

NicolasLM:
The distribution will only include the code that it actually
needed to use the library. Unit tests, live tests and examples
are moved to the root of the repository, like on most other
Python projects.

Fixes #266


HACKING.rst
```diff
@@ -84,5 +84,5 @@ Protocol level unit tests should not act against a real IMAP server
 but should use canned data instead. The IMAPClientTest base class
 should typically be used as the base class for any tests - it provides
 a mock IMAPClient instance at `self.client`. See the tests in
-imapclient/tests/test_imapclient.py for examples of how to write unit
-tests using this approach.
+`tests/test_imapclient.py` for examples of how to write unit tests using
+this approach.
```
mjs: Thank you for catching stuff like this.

setup.py
```diff
@@ -89,7 +90,7 @@ def run_tests(self):
       license="http://en.wikipedia.org/wiki/BSD_licenses",
       url="http://imapclient.freshfoo.com/",
       download_url='http://freshfoo.com/projects/IMAPClient/IMAPClient-%s.zip' % info['version'],
-      packages=find_packages(),
+      packages=['imapclient'],
```
mjs: Good

tests/test_datetime_util.py
```diff
@@ -7,13 +7,13 @@
 from datetime import datetime, date
 
 
-from ..datetime_util import (
+from imapclient.datetime_util import (
```
mjs: Great. I've been meaning to fix all this up.

mjs:
Fabulous. Thank you.

PR #277, opened 2017-08-22
Establish secure connections by default

### LINKED ISSUE #273: Make SSL connections the default
In 2017 there is little reasons to connect to a mailbox over plain text.

As this is a breaking change I propose to do it for 2.0.0.

NicolasLM:
Support for SSL/TLS in both Python and IMAP servers is good enough
to make it a sensible default.

Users can still establish connections over plain text by passing
`ssl=False` when instantiating an `IMAPClient` object.

Fixes #273

### DIFF
--- a/imapclient/imapclient.py
+++ b/imapclient/imapclient.py
@@ -85,16 +85,17 @@ class IMAPClient(object):
     """A connection to the IMAP server specified by *host* is made when
     this class is instantiated.
 
-    *port* defaults to 143, or 993 if *ssl* is ``True``.
+    *port* defaults to 993, or 143 if *ssl* is ``False``.
 
     If *use_uid* is ``True`` unique message UIDs be used for all calls
     that accept message ids (defaults to ``True``).
 
-    If *ssl* is ``True`` an SSL connection will be made (defaults to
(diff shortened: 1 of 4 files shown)


imapclient/imapclient.py
```diff
@@ -132,6 +133,10 @@ def __init__(self, host, port=None, use_uid=True, ssl=False, stream=False,
         elif port is None:
             port = ssl and 993 or 143
 
+        if ssl and port == 143:
+            logger.warning("Attempting to establish an encrypted connection "
```
mlorant: Great warning and debug logs indeed!

imapclient/imapclient.py
```diff
@@ -146,7 +151,8 @@ def __init__(self, host, port=None, use_uid=True, ssl=False, stream=False,
         self._cached_capabilities = None
 
         self._imap = self._create_IMAP4()
-        logger.debug("Connected to host %s", self.host)
+        logger.debug("Connected to host %s over %s", self.host,
+                     "SSL/TLS" if ssl else "plain text")
```
mjs: It might be nice to include the port in this message too

PR #279, opened 2017-08-23
Introduce a context manager that closes opened connections

### LINKED ISSUE #276: Introduce a context manager for IMAPClient instances
A context manager would prevent accidentally leaking sockets when the user forgets to call `logout()`.

NicolasLM:
To prevent users of the library from inadvertently leaking
sockets.

Fixes #276

### DIFF
--- a/doc/src/advanced.rst
+++ b/doc/src/advanced.rst
@@ -3,6 +3,33 @@ Advanced Usage
 This document covers some more advanced features and tips for handling
 specific usages.
 
+Cleaning Up Connections
+~~~~~~~~~~~~~~~~~~~~~~~
+
+To communicate with the server, IMAPClient establishes a TCP connection. It is
+important for long-lived processes to always close connections at some
+point to avoid leaking memory and file descriptors. This is usually done with
+the ``logout`` method::
+
+  import imapclient
+
+  c = imapclient.IMAPClient(host="imap.foo.org")
+  c.login("bar@foo.org", "passwd")
+  c.select_folder("INBOX")
+  c.logout()
+
+However if an error is raised when selecting the folder, the connection may be
+left open.
+
+IMAPClient may be used as a context manager that automatically closes
+connections when they are not needed anymore::
+
+  import imapclient
+
+  with imapclient.IMAPClient(host="imap.foo.org") as c:
+      c.login("bar@foo.org", "passwd")
+      c.select_folder("INBOX")
+
 Watching a mailbox asynchronously using idle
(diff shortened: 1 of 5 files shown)


doc/src/advanced.rst
```diff
@@ -3,6 +3,33 @@ Advanced Usage
...
+
+  c = imapclient.IMAPClient(host="imap.foo.org")
+  c.login("bar@foo.org", "passwd")
+  c.select_folder("INBOX")
+  c.logout()
+
+However if an error is raised when selecting the folder, the connection may be
+left open.
+
+IMAPClient provides a context manager that automatically closes
```
mjs: "IMAPClient may be used as a context manager... " might be slightly better?

imapclient/imapclient.py
```diff
@@ -162,9 +169,22 @@ def __init__(self, host, port=None, use_uid=True, ssl=True, stream=False,
...
 
-        self._idle_tag = None
+    def __enter__(self):
+        return self
 
-        self._set_timeout()
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        """Logout and closes the connection when exiting the context manager.
+
+        All exceptions are caught because an error in `logout` or `shutdown`
```
mjs: "All exceptions during logout and connection shutdown are caught because an error here usually means the connection was already closed."

tests/test_imapclient.py
```diff
@@ -599,3 +599,27 @@ class TestShutdown(IMAPClientTest):
...
+
+        with self.client as client:
+            self.assertIsInstance(client, IMAPClient)
+
+        self.client._imap.logout.assert_called_once_with()
+        self.client._imap.shutdown.assert_called_once_with()
+        mock_logger.info.assert_called_once_with(
+            'Could not close the connection cleanly: %s',
+            self.client._imap.shutdown.side_effect
+        )
```
mjs: Please consider one more test where an exception occurs inside the `with` block. We want to be sure it isn't swallowed by the context manager.

doc/src/advanced.rst
```diff
@@ -3,6 +3,33 @@ Advanced Usage
 This document covers some more advanced features and tips for handling
 specific usages.
 
+Prevent leaving unclosed connections
```
mjs: "Cleaning Up Connections" might be better

tests/test_imapclient.py
NicolasLM: Done

mjs:
Awesome stuff. Thank you!

Issue #36, opened 2010-03-17
UIDPLUS support


mjs:
Originally reported by: **Menno Smits (Bitbucket: [mjs0](https://bitbucket.org/mjs0))**

---

http://tools.ietf.org/html/rfc4315
- Support "UID EXPUNGE" command
- Parsing of APPENDUID response from append()
- Parsing of COPYUID response from copy()
- Test handling of UIDNOTSTICKY response from select_folder()

---
- Bitbucket: https://bitbucket.org/mjs0/imapclient/issue/39

axoroll7:
News about UID EXPUNGE :
issue #496
pull request #508

PR #287, opened 2017-09-18
allow EXPUNGE by ID

davebrown:
Patch to allow EXPUNGE by ID, and avoid the problem where folder-wide expunge deletes all messages.

Implements one item on the list for https://github.com/mjs/imapclient/issues/36

The idea is to overload arguments to expunge(): if messages are specified, do a `'UID EXPUNGE <messages>'`, else it is a folder-wide expunge(). The former won't work if `use_uid=False`. (I don't have insight how many IMAPClient users turn off uid's, and whether this would be better as a different method).

The return type is None when expunge-ing by id. imaplib.py returns `('OK', [None])` from a `UID EXPUNGE`.

### DIFF
--- a/imapclient/imapclient.py
+++ b/imapclient/imapclient.py
@@ -1089,7 +1089,7 @@ class IMAPClient(object):
                                        self._normalise_folder(folder),
                                        uid=True, unpack=True)
 
-    def expunge(self):
+    def expunge(self, messages=None):
         """Remove any messages from the currently selected folder that
         have the ``\\Deleted`` flag set.
 
@@ -1108,6 +1108,10 @@ class IMAPClient(object):
         See :rfc:`3501#section-6.4.3` section 6.4.3 and
         :rfc:`3501#section-7.4.1` section 7.4.1 for more details.
         """
+        if messages:
+          if not self.use_uid:
+            raise ValueError('cannot EXPUNGE by ID when not using uids')
+          return self._command_and_check('EXPUNGE', join_message_ids(messages), uid=True)
         tag = self._imap._command('EXPUNGE')
         return self._consume_until_tagged_response(tag, 'EXPUNGE')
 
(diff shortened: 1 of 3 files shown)


imapclient/imapclient.py
```diff
@@ -1107,7 +1108,21 @@ def expunge(self):
...
+        the ``\\Deleted`` flag set. The return value is ``None`` in
+        this case.
+
+        Expunging messages by id(s) requires that *use_uid* is
+        ``True`` for the client.
+
+        See :rfc:`4315#section-2.1` section 2.1 for more details.
         """
+        if messages:
+          if not self.use_uid:
```
NicolasLM: Could you indent with four spaces?

NicolasLM:
Thanks for your PR, could you try adding a small unit test as well?

imapclient/livetest.py
```diff
@@ -918,6 +918,26 @@ def test_expunge(self):
...
+        def test_id_expunge(self):
+          if not self.client.use_uid:
+            self.skipTest('test instance not configured for UID operations')
+          folder = self.add_prefix_to_folder('test_id_expunge')
+          self.client.create_folder(folder)
+          self.client.select_folder(folder)
+          for i in range(3):
+            self.client.append(folder, 'Subject: msg %d\r\n\r\nbody %d\r\n\r\n' % (i, i))
+          messages = self.client.search()
+          self.assertEquals(len(messages), 3)
```
mlorant: `assertEquals` is a deprecated alias of `assertEqual` (without final s). The others test are using the latter, so it would be better without the extra `s` :) 

And same thing than NicolasLM: four spaces indent on this file as well.

imapclient/imapclient.py
davebrown: done, rev# b566947e46dd5699f66c12507e383f7578f36942

imapclient/livetest.py
davebrown: 4 space indent and `assertEquals` -> `assertEqual` done, rev# b566947e46dd5699f66c12507e383f7578f36942

Issue #285, opened 2017-09-15
Separate connect timeout from read/write timeout


NicolasLM:
Most of the times:
- connecting to a server is a very fast operation (the heavy work is done by the kernel so a slow application does not impact connection much)
- once connected, slow servers can take minutes to execute an IMAP command (extreme example: reading an old email from tape storage)

Thus it makes sense to have a low timeout for connecting and a longer one for read/write operations.

See for instance: http://urllib3.readthedocs.io/en/latest/user-guide.html#using-timeouts

mlorant:
For reference, see ML topic _[Timeout when trying to append a big/long message into a mailbox](https://groups.io/g/imapclient/topic/6016437)_ (main subject not related but the idea pop there)

NicolasLM:
Closed by #289

PR #289, opened 2017-09-25
#285 - Separate connect timeout from read/write timeout

mlorant:
Still a WIP, seems to work on my mailbox, further tests would be appreciated. Things to do:

- [x] Implement something backward-compatible to handle single timeout AND two different timeouts
- [x] Add doc about `SocketTimeout` namedtuple
- [x] Check if all types of IMAP connection (plain, TLS, etc) are handled correctly
- [x] Add tests... How?
- [x] Add release notes

### DIFF
--- a/imapclient/imapclient.py
+++ b/imapclient/imapclient.py
@@ -10,6 +10,7 @@ import select
 import socket
 import sys
 import re
+from collections import namedtuple
 from datetime import datetime, date
 from operator import itemgetter
 from logging import LoggerAdapter, getLogger
@@ -32,7 +33,8 @@ if PY3:
 
 logger = getLogger(__name__)
 
-__all__ = ['IMAPClient', 'DELETED', 'SEEN', 'ANSWERED', 'FLAGGED', 'DRAFT', 'RECENT']
+__all__ = ['IMAPClient', 'SocketTimeout',
+           'DELETED', 'SEEN', 'ANSWERED', 'FLAGGED', 'DRAFT', 'RECENT']
 
 
 # We also offer the gmail-specific XLIST command...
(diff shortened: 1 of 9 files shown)


imapclient/util.py
```diff
@@ -17,3 +18,9 @@ def to_bytes(s, charset='ascii'):
     if isinstance(s, text_type):
         return s.encode(charset)
     return s
+
+
+class SocketTimeout(namedtuple("SocketTimeout", "connect read")):
```
NicolasLM: I am not sure this class really belongs in `util`, I would rather see it in the main `imapclient.imapclient`.

NicolasLM:
Looks good to me 👍 

For unit tests there is not much more you can do than asserting that `create_connection` was called with the connect timeout.

For live tests, you can create one with a ridiculously low connect timeout that will trigger the timeout error all the time.

Don't forget to document the feature in the `IMAPClient` docstring, otherwise people will miss it.

mlorant:
@NicolasLM Thanks for the review. Just one more thing and we're good: do you know how to add a namedtuple with autodoc in Sphinx? I would like to add the doc of the namedtuple in `api.rst` but I did not find anything in the Sphinx doc, except using automodule, but it would not be optimal here...

imapclient/imapclient.py
```diff
@@ -82,6 +83,18 @@ def __new__(cls, personal, other, shared):
     shared = property(itemgetter(2))
 
 
+class SocketTimeout(namedtuple("SocketTimeout", "connect read")):
+    """Represents timeout configuration for an IMAP connection.
+
+    :ivar connect: maximum time to wait for a connection attempt to remote server
+    :ivar read: maximum time to wait for performing a read/write operation
+
+    As an example, ``SocketTimeout(connect=5, read=10)`` will make the socket
```
NicolasLM: Nitpick but I would change the example to `connect=15, read=60` to provide values that are a good default for real life with slow servers.

NicolasLM:
About Sphinx, this works in `api.rst`:
```
.. autoclass:: imapclient.IMAPClient
   :members:

.. autoclass:: imapclient.imapclient.SocketTimeout
   :members:
```

mlorant:
Alright, the real problem was I forgot to include `SocketTimeout` in `__all__` in fact. I tried `.. autoclass:: imapclient.SocketTimeout` before, but was not working because of this mistake.

I'm considering this done -> removing "WIP" tag.

NicolasLM:
Perfect!

PR #291, opened 2017-09-26
Avoid using assert when parsing server response

### LINKED ISSUE #288: Avoid using assert when parsing server response
Some servers are not totally compliant with the IMAP protocol or sometimes give buggy output like: `"TEXT" "PLAIN" ("charset " " "utf-8"")`.

I believe IMAPClient should not `assert` when parsing the response from a server. Assert makes it hard to recover from such errors. A custom IMAPClient exception would be more helpful to the user.

NicolasLM:
To allow users to recover from bad server responses without
catching broad AssertionError, a new exception is introduced.

Fixes #288

### DIFF
--- a/doc/src/releases.rst
+++ b/doc/src/releases.rst
@@ -21,6 +21,8 @@ Changed
   are not used anymore.
 - More precise exceptions available in `imapclient.exceptions` are raised when
   an error happens
+- `imapclient.exceptions.ProtocolError` is now raised when the reply from a
+  remote server violates the IMAP protocol.
 - GMail labels are now strings instead of bytes in Python 3.
 
 Fixed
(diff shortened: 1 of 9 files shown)


NicolasLM:
@mjs There is an undocumented `ParseError` in `imapclient.response_parser`. Should it be replaced by this new error?

mlorant:
Could you add the following test in `test_imapclient.py` ?

```
    def test_tagged_response_with_parse_error(self):
        client = self.client
        client._imap.tagged_commands = {sentinel.tag: None}
        client._imap._get_response = lambda: b'NOT-A-STAR 99 EXISTS'

        with self.assertRaises(ProtocolException):
            client._consume_until_tagged_response(sentinel.tag, b'IDLE')
```
It may not be perfect (With a star, the command does not finish since the response is not complete) but as long as it raises the ProtocolError...

NicolasLM:
Test case added, thanks.

mjs:
@NicolasLM : Yes, please have the parser raise ProtocolError instead of ParseError. I've looked at where ParseError is used and that makes sense.

NicolasLM:
Done.

Issue #290, opened 2017-09-25
AttributeError, 'SSLSocket' object has no attribute '_conn' on logout


M0rtalW0mbat:
Hello, i have some problem during logout process, when working with yandex imap.
I use 32bit python 3.6.2 on 64bit Windows 10, and imapclient 1.0.2

This code is short example for demonstration issue:
context = ssl.create_default_context()
context.check_hostname = False
context.verify_mode = ssl.CERT_NONE
server = imapclient.IMAPClient('imap.yandex.ru', ssl=True, ssl_context=context)
server.debug = 5
server.login('example@yandex.ru', '*******')
server.logout()

Debug traces:
54:58.971300 > b'JNAB1 LOGIN example@yandex.ru "*******"'
54:59.066312 < b'* CAPABILITY IMAP4rev1 CHILDREN UNSELECT LITERAL+ NAMESPACE XLIST BINARY UIDPLUS ENABLE ID IDLE MOVE'
54:59.067313 	matched r'b'\\* (?P<type>[A-Z-]+)( (?P<data>.*))?'' => (b'CAPABILITY', b' IMAP4rev1 CHILDREN UNSELECT LITERAL+ NAMESPACE XLIST BINARY UIDPLUS ENABLE ID IDLE MOVE', b'IMAP4rev1 CHILDREN UNSELECT LITERAL+ NAMESPACE XLIST BINARY UIDPLUS ENABLE ID IDLE MOVE')
54:59.068312 untagged_responses[CAPABILITY] 0 += ["b'IMAP4rev1 CHILDREN UNSELECT LITERAL+ NAMESPACE XLIST BINARY UIDPLUS ENABLE ID IDLE MOVE'"]
54:59.069313 < b'JNAB1 OK LOGIN Completed.'
54:59.070313 	matched r'b'(?P<tag>JNAB\\d+) (?P<type>[A-Z]+) (?P<data>.*)'' => (b'JNAB1', b'OK', b'LOGIN Completed.')
54:59.071313 > b'JNAB2 LOGOUT'
54:59.133324 < b'* BYE IMAP4rev1 Server logging out'
54:59.135324 	matched r'b'\\* (?P<type>[A-Z-]+)( (?P<data>.*))?'' => (b'BYE', b' IMAP4rev1 Server logging out', b'IMAP4rev1 Server logging out')
54:59.135324 untagged_responses[BYE] 0 += ["b'IMAP4rev1 Server logging out'"]
[...]

PR #292, opened 2017-09-26
Fix AttributeError, 'SSLSocket' object has no attribute '_conn' on logout

M0rtalW0mbat:
Dirty fix for #290


M0rtalW0mbat:
backports.ssl.SSLContext must be used!

PR #299, opened 2017-10-05
Remove backports.ssl dependency

mlorant:
As said in #217, `backports.ssl` is no longer needed with Python 2.7+ and 3.4+ and it brings more problems than it resolves. Since it is no longer needed per our support plan, this PR removes any usage of `backports.ssl` + the dependency in setup.py.

### DIFF
--- a/doc/src/concepts.rst
+++ b/doc/src/concepts.rst
@@ -72,29 +72,22 @@ folder names are returned as str (Python 2) or bytes (Python 3).
 
 TLS/SSL
 ~~~~~~~
-IMAPClient established secure connections by default. It uses sensible TLS
-parameters for encrypted connections and also allows for a high level of
-control of TLS parameters if required, with the ``ssl`` module.
+IMAPClient uses sensible TLS parameter defaults for encrypted
+connections and also allows for a high level of control of TLS
+parameters if required. It uses the built-in `ssl` package, 
+provided since Python 2.7.9 and 3.4.
 
-.. note::
+TLS parameters are controlled by passing a ``ssl.SSLContext``
+when creating an IMAPClient instance. When ``ssl=True`` is used
+without passing a SSLContext, a default context is used. The default
+context avoids the use of known insecure ciphers and SSL protocol
+versions, with certificate verification and hostname verification
+turned on. The default context will use system installed certificate
+authority trust chains, if available.
 
-    For older versions of Python which do not provide a recent enough ``ssl``
-    module, IMAPClient installs `backports.ssl
-    <https://github.com/alekstorm/backports.ssl>`_ as a compatibility layer.
-    You are highly encouraged to use a recent version of Python if security is
-    of prime importance for you.
-
-TLS parameters are controlled by passing an ``ssl.SSLContext`` (or
-``backports.ssl.SSLContext`` when appropriate) when creating an IMAPClient
-instance. When ``ssl=True`` is used without passing a SSLContext, a default
-context is used. The default context avoids the use of known insecure ciphers
-and SSL protocol versions, with certificate verification and hostname
-verification turned on. The default context will use system installed
-certificate authority trust chains, if available.
-
-:py:func:`ssl.create_default_context()` returns a safe default context. When
-constructing a custom context it is usually best to start with the default
-context and modify it to suit your needs.
+When constructing a custom context it is usually best to start with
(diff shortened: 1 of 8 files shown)


NicolasLM:
The small problem is that the modern `ssl` module is available from Python 2.7.9+.

`backports.ssl` is so broken anyway that I personally don't mind starting the compatibility of IMAPClient `2.0` at 2.7.9 instead of 2.7.0 as it is currently on master.

NicolasLM:
@mjs What do you think?

mjs:
I'm OK with promising full TLS support with Python 2.7.9 and up, but it would be good if IMAPClient could at least offer TLS - albeit not as secure or configurable as it could be - with 2.7.0 and up. I know there are some big users of IMAPClient on 2.7.{0-8} .

imapclient/tls.py
```diff
@@ -5,200 +5,30 @@
...
-
-        def __init__(self, sslsock):
-            self.sslsock = sslsock
+def create_default_context(cafile=None, capath=None, cadata=None):
+    """TODO: doc
+    """
+    return ssl.create_default_context(purpose=ssl.Purpose.CLIENT_AUTH,
+                                      cafile=cafile,
+                                      capath=capath,
+                                      cadata=cadata)
```
mjs: We should make sure that this context actually works with the big IMAP providers. TLS provided by IMAP servers is usually somewhat behind what is offered by web servers in terms of TLS versions, ciphers etc.

imapclient/tls.py
```diff
@@ -5,200 +5,30 @@
...
-    wrap_socket = lambda sock, context, host: _wrap_socket(sock, context, host)
-
-
-    # TODO: get shutdown added in backports.ssl.SSLSocket
-    class _SSLSocketWithShutdown(object):
-
-        def __init__(self, sslsock):
-            self.sslsock = sslsock
+def create_default_context(cafile=None, capath=None, cadata=None):
+    """TODO: doc
```
mjs: Needs doing

doc/src/concepts.rst
```diff
@@ -72,29 +72,22 @@ folder names are returned as str (Python 2) or bytes (Python 3).
...
-verification turned on. The default context will use system installed
-certificate authority trust chains, if available.
-
-:py:func:`ssl.create_default_context()` returns a safe default context. When
-constructing a custom context it is usually best to start with the default
-context and modify it to suit your needs.
+IMAPClient uses sensible TLS parameter defaults for encrypted
+connections and also allows for a high level of control of TLS
+parameters if required. It uses the built-in `ssl` package, 
+provided since Python 3.4.
```
mjs: ... since Python 2.7.9 and Python 3.4.

imapclient/tls.py
NicolasLM: We use the default context to connect to the big players without any issue.

imapclient/tls.py
NicolasLM: Do we need this function at all?

mlorant:
I don't know why but I did not notice @mjs's comments before... Maybe I've deleted the notification too fast. Will fix this asap.

imapclient/tls.py
mlorant: It was useful before to use the backports solution if needed, but now we can use `ssl` module everywhere, right?

NicolasLM: Yes, it looks more explicit to use `ssl` directly. It can go away unless we want to keep backward compatibility.

imapclient/tls.py
mjs: Ok, that's great. 

The only reason to keep this is as a convenience so you don't have to set the CLIENT_AUTH purpose, but that's probably not reason enough to keep it. I'm happy for it to go.

There may be examples in the docs which need to be updated.

mlorant:
Oddly, examples were not using `imapclient.tls` but the `ssl` one. Still, I removed the reference in the doc, suggesting to create a default context directly with the `ssl` module if someone needs it.

NicolasLM:
I did a pass a few weeks ago on the examples. Since `master` could work without `backports.ssl` I thought it was better to encourage people to use `ssl` directly.

NicolasLM:
`backports.ssl` is dead, long live `ssl` 🎉

mjs:
@arnisoph Great! Thanks for reporting back.

PR #300, opened 2017-10-09
Refactor imap_utf7 module to make it easier to understand and fix #187

### LINKED ISSUE #187: Investigate claims that UTF-7 encoding/decoding doesn't always work

---

See:  https://github.com/MarechJ/py3_imap_utf7

If/when any problems are fixed, ask the author to update the page. Also add comments to SO where this is mentioned.

---
- Bitbucket: https://bitbucket.org/mjs0/imapclient/issue/190

mlorant:
Instead of trying to fix the existing code for the encoder (the part that was buggy in issue #187), I took the RFC and follow each step to reproduce the algorithm. The initial code was almost correct (I guess the problem came from the missing `rstrip` I add in `base64_utf7_encode`) but anyway, my version has better variable names and comments... 😅

I've also renamed variables in the decode function to make it consistent + easier to read later. Also added comments to explain what it does.

Finally, I added the use case that was failing in the issue to validate it. If you see any other cases failing with the 1.0.2 implementation and that would be interesting for the non-regression test, please let me know.

### DIFF
--- a/doc/src/releases.rst
+++ b/doc/src/releases.rst
@@ -6,7 +6,7 @@
 
 Added
 -----
-- Connection and read/write operations timeout can now be distinct, 
+- Connection and read/write operations timeout can now be distinct,
   using `imapclient.SocketTimeout` namedtuple as `timeout` parameter.
 - A context manager is introduced to automatically close connections to remote
   servers.
@@ -19,6 +19,12 @@ Changed
   are not used anymore.
 - More precise exceptions available in `imapclient.exceptions` are raised when
   an error happens
+
+Fixed
+-----
+- Modified UTF-7 encoding function had quirks in its original algorithm,
+  leading to incorrect encoded output in some cases. The algorithm, described
+  in RFC 3501, has been reimplemented to fix #187 and is better documented.
 
 Other
 -----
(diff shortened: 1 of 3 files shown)


NicolasLM:
Awesome! Can you add a changelog entry?

mlorant:
Changelog added (in 2.0.0, new section "Fixed")

doc/src/releases.rst
```diff
@@ -20,6 +20,12 @@ Changed
 - More precise exceptions available in `imapclient.exceptions` are raised when
   an error happens
 
+Fixed
+-----
+- Modified UTF-7 encoding function had quirks in its original algorithm,
+  leading to misencoded bytes data in some cases. The algorithm, described
```
mjs: s/misencoded/incorrectly encoded/

doc/src/releases.rst
```diff
@@ -20,6 +20,12 @@ Changed
...
+  in RFC 3501, have been reimplemented to fix #187 and is better documented.
```
mjs: s/have/has/

doc/src/releases.rst
mlorant: Both amended, thanks!

mjs:
Build failure appears spurious. Re-running the failed Python 3.6 build to be sure.

mjs:
All good.

PR #306, opened 2017-10-29
quote empty strings as otherwise they would be ignored by the server

jonasc:
I've run into a problem when using the search criterion `HEADER`. As [RFC 3501](https://tools.ietf.org/html/rfc3501#section-6.4.4) tells us the syntax is `HEADER <field-name> <string>`. It also says: " If the string to search is zero-length, this matches all messages that have a header line with the specified field-name regardless of the contents."
When you issue `server.search(['HEADER', 'List-Id', ''])` the zero-length argument is ignored resulting in a syntax error.

This pull request fixes this problem. I do not know whether this is the right place to fix this problem but if so I would be glad if the PR could be merged.

### DIFF
--- a/tests/test_search.py
+++ b/tests/test_search.py
@@ -99,6 +99,10 @@ class TestSearch(TestSearchBase):
         self.client.search(['NOT', ('SUBJECT', 'topic',  'TO', 'some@email.com')])
         self.check_call([b'NOT', b'(SUBJECT', b'topic', b'TO', b'some@email.com)'])
 
+    def test_quote_empty_strings(self):
+        self.client.search(['HEADER', 'List-Id', ''])
+        self.check_call([b'HEADER', b'List-Id', b'""'])
+
     def test_search_custom_exception_with_invalid_list(self):
         def search_bad_command_exp(*args, **kwargs):
             raise imaplib.IMAP4.error('SEARCH command error: BAD ["Unknown argument NOT DELETED"]')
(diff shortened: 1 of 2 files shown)


NicolasLM:
Hey @jonasc, thanks for your contribution, it looks good to me.

Two things could be improved:
* Adding a small unit-test to make sure that we don't break this behavior during a potential refactoring
* Rework a bit the commit message to explain the "why" instead of the "what" (basically the description of the pull request that you wrote)

mjs:
Thanks for the fix. :+1: to @NicolasLM's comments.

jonasc:
I rephrased the commit message and added a small test case. I hope it's better now, if not just let me know!

tests/test_search.py
```diff
@@ -99,6 +99,10 @@ def test_nested_tuple(self):
         self.client.search(['NOT', ('SUBJECT', 'topic',  'TO', 'some@email.com')])
         self.check_call([b'NOT', b'(SUBJECT', b'topic', b'TO', b'some@email.com)'])
 
+    def test_quote_empty_strings(self):
```
mlorant: I would either merge this test into `test_quoting` (line 63) or place it just under, as they are related.

By the way I am wondering why there is no tests to check the escape of backslashes, but this is not in the scope of your PR :)

mlorant:
Just one comment from my perfectionist point of view but otherwise it's all good for me, thank you!

tests/test_search.py
NicolasLM: 👍

jonasc:
Just figured out (see https://travis-ci.org/mjs/imapclient/builds/295220294) that merging two test cases into one function does not work. Thus the last commit ed70c33.

NicolasLM:
@jonasc thank you for your contribution!

PR #307, opened 2017-10-30
Ignore invalid encoding in Address string representation

### LINKED ISSUE #243: `_convert_to_ENVELOPE` (and subcall to `util.to_unicode`) only support ascii data
When using `fetch(uids, ['ENVELOPE'])`, we get a `imapclient.response_types.Envelope` namedtuple. However, this object may result in subcomponent that call `imapclient.util.to_unicode`. However this methods, decode bytes data only with the ascii codec.

For some reasons, some senders does not encode the data before sending the e-mail, resulting in object like this...

```In [5]: addr = Address(b'Ren\xe9', None, 'rene', 'domain.com')
In [6]: str(addr)
---------------------------------------------------------------------------
/dir/to/imapclient/response_types.py in __str__(self)
     56     def __str__(self):
     57         return formataddr((
---> 58             to_unicode(self.name),
     59             to_unicode(self.mailbox) + '@' + to_unicode(self.host)))
     60 

/dir/to/imapclient/util.py in to_unicode(s)
     10 def to_unicode(s):
     11     if isinstance(s, binary_type):
---> 12         return s.decode('ascii')
     13     return s
     14 

UnicodeDecodeError: 'ascii' codec can't decode byte 0xe9 in position 3: ordinal not in range(128)
```

I think `imapclient` should handle this and either raise a proper custom exception which informs the developer why the data is incorrect or decode the data (using latin1 codec in this case), if possible...

mlorant:
Fixes #243

### DIFF
--- a/imapclient/util.py
+++ b/imapclient/util.py
@@ -4,12 +4,22 @@
 
 from __future__ import unicode_literals
 
+import logging
 from six import binary_type, text_type
+
+logger = logging.getLogger(__name__)
 
 
 def to_unicode(s):
     if isinstance(s, binary_type):
-        return s.decode('ascii')
+        try:
+            return s.decode('ascii')
+        except UnicodeDecodeError:
+            logger.warning(
+                "An error occurred while decoding %s in ASCII 'strict' mode. Fallback to "
+                "'ignore' errors handling, some characters might have been stripped", s
+            )
+            return s.decode('ascii', 'ignore')
     return s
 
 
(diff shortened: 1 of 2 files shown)


NicolasLM:
It's probably the way to go but I'm a bit reluctant to let things be silently ignored. Could we log a warning?

mlorant:
Indeed, I can use a `try/except`, issuing a warning and using `.encode`  again but with `'ignore` in case of exception.

NicolasLM:
Perfect 👌

mlorant:
Should also be merged in 1.1.0 (see milestone on #243) :)

NicolasLM:
#306 should also be backported to `1.1.0`. We don't have a handy place to write this down 😞

PR #309, opened 2017-11-09
Use cryptography < 2.0 on Python 3.3

### LINKED ISSUE #305: cryptography no longer works with Python 3.3
cryptography 2.0 doesn't work with Python 3.3 so we need to tweak our dependencies in setup.py.
This only needs to be done in the 1.x branch.

See here for how it was handled for another project: https://gerrit.wikimedia.org/r/#/c/384321/2/setup.py

NicolasLM:
The newest version of cryptography dropped support for Python 3.3.

Fixes #305


NicolasLM:
That's possible but it would prevent any user of Python 3.4+ from using cryptography 2.0. In my opinion this is a problem as this library is about security. Who knows if they will backport security fixes to their 1.x branch.

mjs:
Ok fair enough. We will just have to remember to not release 1.x as a wheel.

PR #310, opened 2017-11-09
Allow old version of Python 2.7 to use TLS

### LINKED ISSUE #301: Make TLS work for Python 2.7 before 2.7.9
Decent TLS support didn't arrive until 2.7.9. A number of big consumers of IMAPClient are on 2.7.{0-8} so it would be helpful to offer some level of TLS support in these versions.

This continues on from #299.

NicolasLM:
The modern implementation of ssl using SSLContext appeared in
Python 2.7.9. Previous versions can use encrypted connections but
cannot precisely configure it.

Fixes #301

### DIFF
--- a/doc/src/concepts.rst
+++ b/doc/src/concepts.rst
@@ -89,6 +89,13 @@ When constructing a custom context it is usually best to start with
 the default context, created by the ``ssl`` module, and modify it to
 suit your needs.
 
+.. warning::
+
+  Users of Python 2.7.0 - 2.7.8 can use TLS but cannot configure
+  the settings via an ``ssl.SSLContext``. These Python versions are
+  also not capable of proper certification verification. It is highly
+  encouraged to upgrade to a more recent version of Python.
+
 The following example shows how to to disable certification
 verification and certificate host name checks if required.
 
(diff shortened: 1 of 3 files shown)


doc/src/concepts.rst
```diff
@@ -89,6 +89,12 @@ When constructing a custom context it is usually best to start with
 the default context, created by the ``ssl`` module, and modify it to
 suit your needs.
 
+.. warning::
+
+  Users of Python 2.7.0 - 2.7.8 can use TLS but cannot precisely configure
```
mjs: s/ precisely//

doc/src/concepts.rst
```diff
@@ -89,6 +89,12 @@ When constructing a custom context it is usually best to start with
...
+  the settings via an ``ssl.SSLContext``, they are highly encouraged to
```
mjs: ...ssl.SSLContext. These Python versions are also not capable of proper certification verification. It is highly encouraged to upgrade to a more recent version of Python.

doc/src/concepts.rst
NicolasLM: All done.

Issue #312, opened 2017-11-13
Remove the use of imaplib


mjs:
Much of IMAPClient could be so much cleaner if it didn't rely on imaplib underneath.

equaeghe:
Do you want to move to an alternative or create your own imaplib replacement. In case of the former, perhaps you are not aware of [aioimaplib](https://github.com/bamthomas/aioimaplib), but the fact that it is GPL-3 may be an obstacle.

mjs:
@equaeghe The plan is to just have a small core of what imaplib does implemented as part of IMAPClient. There's really not that much required that IMAPClient doesn't have already.

PR #314, opened 2017-11-13
Use imaplib's starttls if available

boehlke:


boehlke:
Sorry for opening another PR... little fuckup, will improve, promise :-)

NicolasLM:
Thank you for your contribution, but I don't think that this change is necessary:
- As mentioned in #312, IMAPClient plans on ditching `imaplib` completely
- As far as I know our own `starttls` works on all Python versions supported (2.7, 3.4, 3.5, 3.6)
- Our own `starttls` and the one from `imaplib` share almost the same code

boehlke:
@NicolasLM Fine, thanks for considering.

PR #315, opened 2017-11-14
Simplify setup.py

NicolasLM:
* Drop ez_setup, it is very unlikely someone still uses it
* Only install `mock` to run tests on Python 2.7
* Don't override unittest discovery command


NicolasLM:
@mjs Maybe a slight modification is needed when building the documentation:

    pip install .[doc]

mlorant:
That sounds right, I haven't pull and tested myself it though.

mjs:
Looks great. `doc/src/contributing.rst` needs to be updated as part of this as well as it mentions `python setup.py test`.

NicolasLM:
@mjs I removed a few things in `doc/src/contributing.rst` that are no longer relevant, but the prefered way to run tests is still `python setup.py test` (see `.travis-ci.yml`).

mjs:
FWIW, `pip install .[doc]` does indeed work. This is mentioned in the contributing docs now.

PR #316, opened 2017-11-15
Handle GMail labels with international characters

### LINKED ISSUE #181: UTF-7 decoding of Gmail labels

---

As worked around here in the Nylas sync engine:
https://github.com/nylas/sync-engine/commit/fdeee59be8feea1c4fac4d9972b3462d128c25f2

Fixing this might break backwards compatibility (so might require a 2.0)

---
- Bitbucket: https://bitbucket.org/mjs0/imapclient/issue/184

NicolasLM:
Encode/decode GMail labels using UTF-7 so that international
characters are handled correctly.

This has the effect of slightly changing IMAPClient API as labels
are now str in Python 3 and unicode in Python 2.7.

Fixes #181

### DIFF
--- a/doc/src/releases.rst
+++ b/doc/src/releases.rst
@@ -19,6 +19,11 @@ Changed
   are not used anymore.
 - More precise exceptions available in `imapclient.exceptions` are raised when
   an error happens
+- GMail labels are now strings instead of bytes in Python 3.
+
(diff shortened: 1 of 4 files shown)


NicolasLM:
@mjs @mlorant note that flags are still bytes.

PR #324, opened 2017-12-15
WIP: Asynchronously wait for IMAP IDLE notifications

NicolasLM:


NicolasLM:
@dsanghan @mjs @mlorant Here is a port of the code we use to listen for IDLE notification from thousands of clients concurrently.

It misses a few things:
- Python 2.7 support, I don't know how important this is since we won't be supporting it forever
- Tests obviously
- Documentation
- Proper IMAP parsing of responses
- Metrics, I had to remove Prometheus metrics because not anyone is interested in them but I need to find a way to optionally plug them back (signals maybe?)

NicolasLM:
Closing this one, while it works and is useful to us, a more generic approach to asynchronous operations should be used instead.

nesttle:
Hi @NicolasLM is there any possibility to use this example with the current version of IMAPClient?

We need to connect to several accounts at the same time, and by default the library does not allow it. Thank you

NicolasLM:
I don't think IMAPClient changed significantly enough to break it, however this code was never merged so you will need to iron out the rough edges.

PR #342, opened 2018-01-18
Add method to fetch Python email Message

### LINKED ISSUE #340: Fetch and parse message as an `email.message.Message`
Python has a convenient type to represent email messages. We could add an alternate `fetch` method that returns it.

Actually there are two types, the current [EmailMessage](https://docs.python.org/3/library/email.message.html#email.message.EmailMessage) and the legacy [Message](https://docs.python.org/3/library/email.compat32-message.html#email.message.Message).

NicolasLM:
Fixes #340

### DIFF
--- a/imapclient/imapclient.py
+++ b/imapclient/imapclient.py
@@ -4,6 +4,7 @@
 
 from __future__ import unicode_literals
 
+import email
 import functools
 import imaplib
 import itertools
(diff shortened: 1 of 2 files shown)


imapclient/imapclient.py
```diff
@@ -1206,6 +1207,38 @@ def fetch(self, messages, data, modifiers=None):
         typ, data = self._imap._untagged_response(typ, data, 'FETCH')
         return parse_fetch_response(data, self.normalise_times, self.use_uid)
 
+    def fetch_email_messages(self, messages, _class=None):
```
mjs: Maybe just call it `fetch_messages`?

imapclient/imapclient.py
```diff
@@ -1206,6 +1207,38 @@ def fetch(self, messages, data, modifiers=None):
...
+        """Retrieve messages as Python email Messages.
+
+        *messages* is a list of message IDs to retrieve.
+
+        The optional *_class* is passed to the `email.parser.Parser` that
+        generates the Message.
+
+        A dictionary is returned, indexed by message number. Values are
+        instances of `email.message.Message` or `email.message.EmailMessage`
+        depending on the *_class* used and the version of Python.
```
mjs: How about:
A dictionary is returned, indexed by message id. By default, the values are instances of `email.message.Message` or `email.message.EmailMessage` depending on the version of Python. If *_class* was specified then values will be instances of that class.

NicolasLM:
After reading comments in #340 and rethinking this through I think it's better to document how to work efficiently with data returned by IMAPClient. This is because the previous (2.7) and new (3.5) `email` package have a lot of different settings and a one size fits all just won't cut it.

PR #350, opened 2018-04-20
Fix error when INTERNALDATE is NIL

carsonip:
NIL is already converted to None during parsing.
The .upper() call on None will cause AttributeError. Catch this case
early and return None early.


NicolasLM:
Looks good, thanks!

PR #361, opened 2018-08-30
Document the recommended way of parsing emails

### LINKED ISSUE #340: Fetch and parse message as an `email.message.Message`

NicolasLM:
Fixes #340


examples/email_parsing.py
```diff
@@ -0,0 +1,17 @@
...
+PASSWORD = 'secret'
+
+with IMAPClient(HOST) as server:
+    server.login(USERNAME, PASSWORD)
+    server.select_folder('INBOX', readonly=True)
+
+    messages = server.search('UNSEEN')
+    for uid, message_content in server.fetch(messages, 'RFC822').items():
+        email_message = email.message_from_bytes(message_content[b'RFC822'])
+        print(email_message.get('From'), email_message.get('Subject'))
```
mjs: Perhaps include the uid in the output?

examples/email_parsing.py
```diff
@@ -0,0 +1,17 @@
...
+HOST = 'imap.host.com'
+USERNAME = 'someuser'
+PASSWORD = 'secret'
+
+with IMAPClient(HOST) as server:
+    server.login(USERNAME, PASSWORD)
+    server.select_folder('INBOX', readonly=True)
+
+    messages = server.search('UNSEEN')
+    for uid, message_content in server.fetch(messages, 'RFC822').items():
```
mjs: message_data instead of message_content?

doc/src/concepts.rst
```diff
@@ -70,6 +70,18 @@ If *folder_encode* is True, all folder names returned by IMAPClient
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+The IMAP protocol gives access to a limited amount of information about emails
+stored on the server. In depth analysis of a message usually requires
+downloading the full message and parsing its content.
+
+The `email <https://docs.python.org/3/library/email.html>`_ package of the
+Python standard library provides a reliable way to transform a raw email into
+a convenient object.
+
+.. literalinclude:: ../../examples/email_parsing.py
```
mjs: This is great. Thanks!

PR #363, opened 2018-09-18
fix find_special_folder

Unpublished:


NicolasLM:
Hi,
Thank you for the patch. It's great that you added a test case, however would you mind describing briefly the problem you're fixing in the commit message? You can amend your commit. Thanks!

Unpublished:
Hey,
sorry for the incomplete PR, I just wanted to quickly fix it and share it. The problem is that previous implementation of `find_special_folder` expects the special flag at the second position, however that's not true for all servers. I fixed that by scanning through all flags.

I'll amend the message tomorrow.

Unpublished:
While updating I noticed another theoretical cornercase (only a single flag), let me know if you still need additional information.

NicolasLM:
Great, thank you!

PR #371, opened 2019-03-19
Catch the right exception in folder_status

carsonip:
Not sure why are we catching ValueError here. I don't think it will ever raise ValueError just by assigning list/tuple's index -1 item to a variable. The exception I got from `response[-1]` is IndexError when the response is an empty list. But if we catch IndexError, `str(response)` is certainly empty list/tuple. Do we still want that exception message?

### DIFF
--- a/imapclient/imapclient.py
+++ b/imapclient/imapclient.py
@@ -848,10 +848,7 @@ class IMAPClient(object):
         fname = self._normalise_folder(folder)
         data = self._command_and_check('status', fname, what_)
         response = parse_response(data)
-        try:
-            status_items = response[-1]
-        except IndexError:
-            raise Exception('folder_status IndexError: Input ' + str(response))
+        status_items = response[-1]
         return dict(as_pairs(status_items))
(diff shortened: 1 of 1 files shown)


NicolasLM:
It looks like the whole try/except makes little sense: by throwing a bare `Exception` is makes the `ValueError`/`IndexError` even less useful.

In my opinion we should either remove it or handle correctly the case of `response` being an empty list.

carsonip:
> It looks like the whole try/except makes little sense: by throwing a bare `Exception` is makes the `ValueError`/`IndexError` even less useful.
> 
> In my opinion we should either remove it or handle correctly the case of `response` being an empty list.

That's very true. Actually my first change was to remove the whole try catch, but I thought there was a reason for the `Exception`. I would go for "removing the try catch". imapclient throws exceptions everywhere when the response is not compliant anyway.

carsonip:
Please squash merge if possible so we don't get 2 small commits.

PR #373, opened 2019-03-25
Faster utf7 encode

carsonip:
~40% faster for input with a mix of unicode and ascii chars. The improvement is more significant for pure ascii chars.

This is achieved by removing the .encode() in the most common path (ascii char) and comparing ordinal values instead of unicode objects.

Actually we can get 10% more by eliminating dot notation, 10-20% more by inlining `consume_b64_buffer` function, some more by caching `ord` global function, but then the code will get quite ugly and I'm not sure if that's welcomed.

### DIFF
--- a/imapclient/imap_utf7.py
+++ b/imapclient/imap_utf7.py
@@ -21,26 +21,29 @@ def encode(s):
     if not isinstance(s, text_type):
         return s
 
-    res = []
+    res = bytearray()
+
     b64_buffer = []
+
     def consume_b64_buffer(buf):
         """
         Consume the buffer by encoding it into a modified base 64 representation
         and surround it with shift characters & and -
         """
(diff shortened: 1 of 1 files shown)


carsonip:
As a side note, combining with Cython, this can achieve a total of 10x improvement.

carsonip:
Forgot to attach the benchmark script and results:
```
# -*- coding: utf-8 -*-
import time
from imapclient.imap_utf7 import encode, decode

q = u'你好嗎你好嗎ab 123&123 abc你' * 50

start = time.time()
for i in xrange(5000):
   encode(q)
print(time.time() - start)
```

CPython2.7
before: 2.45s
after: 1.48s

PyPy2.7 v7.1
before: 0.284s
after: 0.198s

PR #377, opened 2019-04-20
poll() when available to surpass 1024 file descriptor limit with select()

jhatch28:
We've been seeing errors from the `select()` system call because we use more than 1,024 file descriptors in our processes. This patches the `ImapClient` to use the `poll()` system call when available.

### DIFF
--- a/imapclient/imapclient.py
+++ b/imapclient/imapclient.py
@@ -28,6 +28,13 @@ from .response_parser import parse_response, parse_message_list, parse_fetch_res
 from .util import to_bytes, to_unicode, assert_imap_protocol, chunk
 xrange = moves.xrange
 
+try:
+    from select import poll
+    POLL_SUPPORT = True
+except:
+    # Fallback to select() on systems that don't support poll()
+    POLL_SUPPORT = False
+
 if PY3:
     long = int  # long is just int in python3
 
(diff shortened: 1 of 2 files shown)


mjs:
Thanks for this. I'm definitely willing to take this change but there seems to be some bits missing. I don't see definitions `POLL_SUPPORT`, `self._poll_socket` and `self._select_poll_socket`.

If you can get this sorted out I'm sure we can get this landed.

jhatch28:
Thanks for the quick reply, looks like I had some confusion on my end missing a commit somehow. Added it so the tests should rerun. 

Let me know if there's anything you'd like to change before merging. I just needed to get a quick patch up to fix the error for us.

imapclient/imapclient.py
```diff
@@ -761,6 +768,26 @@ def idle(self):
...
+        return poller.poll(timeout)
+
+    def _select_poll_socket(self, sock, timeout=None):
+        """
+        Polls the socket for events telling us it's available to read.
+        This implementation is a fallback because it FAILS if your process
+        has more than 1024 file descriptors.
+        We still need this for Windows and some other niche systems.
+        """
+        return select.select([sock], [], [], timeout)[0]
```
mjs: Would you mind moving `_poll_socket` and `_select_poll_socket` out to standalone funcs in a new module (`poll.py`?). This module could then have a `get_poll_func` function which returns the appropriate variant depending on what's available. 

This module should have it's own unit tests.

This would be a little cleaner and will probably make the tests simpler too (just patch `get_poll_func`).

PR #382, opened 2019-07-09
Add ESORT/ESEARCH support.

jap:
ESORT and ESEARCH are extensions to SORT and SEARCH allowing users to
also get the number of messages (COUNT) matching their criteria, the
MIN and MAX ids of those messages, and to also only return part of the
message ids for more efficient pagination.


imapclient/response_parser.py
```diff
@@ -152,6 +152,69 @@ def parse_fetch_response(text, normalise_times=True, uid_is_key=True):
...
+        while 1:
+            bite = six.next(it)
+            if isinstance(bite, tuple) and len(bite) == 2 and bite[0] == b'TAG':
+                # FIXME: should verify we only consume messages matching our tag
+                continue
+            elif bite == b'UID':  # this is just a marker that we are using UIDs, ignore it.
+                continue
+            elif bite == b'ALL':
+                message_bite = six.next(it)
+                retval[bite + b'_RAW'] = _raw_as_bytes(message_bite)
```
mjs: Why is the raw value being returned along with the parsed value? Is the parsed version not more useful?

imapclient/response_parser.py
```diff
@@ -152,6 +152,69 @@ def parse_fetch_response(text, normalise_times=True, uid_is_key=True):
...
+                retval[bite] = _parse_compact_message_list(message_bite)
+            elif bite == b'PARTIAL':
+                message_bite = six.next(it)[1]
+                retval[bite + b'_RAW'] = _raw_as_bytes(message_bite)
```
mjs: and again

imapclient/response_parser.py
jap: It should be in the docstring - the raw version allows for a small optimisation, as it can be passed on to a subsequent fetch as well (removing the need to serialise again, and the raw version might contain ranges, making for a small size).
I'm not sure if it has any real life benefits, as the serialisation and transport costs will probably be dwarfed by the time it takes the server to construct and send the response, so if you feel it makes the interface too complicated, I can remove it as well. Real life data point: we skipped using the raw version in the end.

mjs: What about making each of these values a (to be implemented) "SequenceSet" object which implement __iter__ to yield the expanded message ids and __str__ to provide the original raw set string?

This makes the raw and parsed versions easily available without requiring separate dictionary values for the raw and parsed variants. Yielding message ids instead of generating a list helps to avoid unnecessary memory consumption in the case of large message ranges.

jap: That sounds like a better way to do this, let me find some time to implement that. Besides `__str__` I'd prefer `__bytes__` as well, except that is a python3 thing, but I can do both. I'll also just fix the `imapclient.util.to_bytes` to check if `__bytes__` is available and use it.

PR #395, opened 2020-02-11
Fix useless ref cycle in lexer

carsonip:
Remove a useless ref cycle in LiteralHandlingIter. This helps the ref count based garbage collection in CPython.


NicolasLM:
Looks good, thank you!

PR #399, opened 2020-05-18
MULTIAPPEND and LITERAL+ support

### LINKED ISSUE #195: MULTIAPPEND support
Originally reported by: **Erik Quaeghebeur (Bitbucket: [equaeghe](https://bitbucket.org/equaeghe))**

---

See https://tools.ietf.org/html/rfc3502 (and perhaps updates).

It would seem impossible to do this by extending the append method without breaking backwards compatability, so perhaps an option is to add a multiappend method taking an iterator over {'msg': ...[, 'flags': ...][, 'date': ...]}-dicts that can fall-back to multiple appends.

This kind of functionality may be useful for clients also working in disconnected mode.

---
- Bitbucket: https://bitbucket.org/mjs0/imapclient/issue/198

akvadrako:
With normal APPEND I can upload about 10 msgs/second. With MULTIAPPEND and LITERAL+ I can get about 600 msgs/second. This solves #195.

I am not sure about the  commented out lines - they are part of imaplib's `append` method, but my testing seemed to work without them.

I added the `_literal` object because it seemed that messages should always be sent as literals, even if they are not 8bit.

BTW, great library. One thing I didn't understand once I got hacking on it is the dependency on imaplib. It doesn't seem to add much but makes half the code use a different execution path.

Closes #195 .

### DIFF
--- a/imapclient/imapclient.py
+++ b/imapclient/imapclient.py
@@ -1290,17 +1290,13 @@ class IMAPClient(object):
 
     @require_capability('MULTIAPPEND')
     def multiappend(self, folder, msgs):
-        """Append messages to *folder* using the MULTIAPPEND feature from RFC3502.
+        """Append messages to *folder* using the MULTIAPPEND feature from :rfc:`3502`.
 
-        *msgs* should be a list of string contains the full message including
+        *msgs* should be a list of strings containing the full message including
         headers.
 
         Returns the APPEND response from the server.
         """
-        # literal = imaplib.MapCRLF.sub(CRLF, message)
-        # if self._imap.utf8_enabled:
-        #    literal = b'UTF8 (' + literal + b')'
-
         msgs = [_literal(to_bytes(m)) for m in msgs]
 
         return self._raw_command(
--- a/imapclient/imapclient.py
+++ b/imapclient/imapclient.py
@@ -1694,7 +1694,7 @@ def _normalise_sort_criteria(criteria, charset=None):
     return b'(' + b' '.join(to_bytes(item).upper() for item in criteria) + b')'
 
 class _literal(bytes):
-    """Hold message data that should always be send as a literal."""
+    """Hold message data that should always be sent as a literal."""
     pass
 
 class _quoted(binary_type):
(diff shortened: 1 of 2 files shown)


imapclient/imapclient.py
```diff
@@ -1288,6 +1288,27 @@ def append(self, folder, msg, flags=(), msg_time=None):
                                        to_bytes(msg),
                                        unpack=True)
 
+    @require_capability('MULTIAPPEND')
+    def multiappend(self, folder, msgs):
+        """Append messages to *folder* using the MULTIAPPEND feature from RFC3502.
```
mjs, line 1293: Please make this an RFC link like this:
```
:rfc:`3502` 
```

imapclient/imapclient.py
```diff
@@ -1288,6 +1288,27 @@ def append(self, folder, msg, flags=(), msg_time=None):
...
+
+        *msgs* should be a list of string contains the full message including
```
mjs, line 1295: `strings containing`

imapclient/imapclient.py
```diff
@@ -1288,6 +1288,27 @@ def append(self, folder, msg, flags=(), msg_time=None):
...
+        headers.
+
+        Returns the APPEND response from the server.
+        """
+        # literal = imaplib.MapCRLF.sub(CRLF, message)
+        # if self._imap.utf8_enabled:
+        #    literal = b'UTF8 (' + literal + b')'
```
mjs, line 1302: I guess just leave this out if it's not needed.

mjs:
> BTW, great library. One thing I didn't understand once I got hacking on it is the dependency on imaplib. It doesn't seem to add much but makes half the code use a different execution path.

IMAPClient started out as a fairly thin wrapper around imaplib which just smoothed over some wrinkles but as the project became more ambitious, it was necessary to work around imaplib. I'd love to see the imaplib dependency removed and I can see a path to doing this but I don't really have the time and motivation to do it all myself. It would certainly allow IMAPClient to be simpler and cleaner in parts.

akvadrako:
Thanks for the quick takeup; I've applied your suggestions.

mjs:
Thanks again for this!

PR #400, opened 2020-05-22
remove imaplib dependency

akvadrako:
I felt like taking a stab at this. It seems to work, though I've only tested against Dovecot. I added a `livetest.sh` file to automate this with podman.

Basically, I incorporated all the calls from imaplib that imapclient was using into IMAPClient or a new conn.IMAP4 class for the low-level stuff. I divided `_command_and_check` into 2 versions:

1. `_uid_command_and_check` -> just calls `_simple_command`, optionally with UID prefix.
2. `_subcommand_and_check` -> wraps an old imaplib method. No UID support.

It should be easy to do a bit more refactoring on a command-by-command basis, merging the old _cmd_XXX() call (that's from imaplib) into the imapclient parent method. Most of the work might actually be updating the unit tests.

There are a few more things to note. It seems like `ACCEPT=UTF8` support was broken, since `imaplib.IMAP4.enable()` wasn't be called - that's where the magic happens. So I just removed that stuff for now. Also, there were a few calls imaplib supported that client doesn't - like get/set annotations. I left those in as private methods.

I'm open to any feedback.

### DIFF
--- a/imapclient/conn.py
+++ b/imapclient/conn.py
@@ -1,0 +1,213 @@
+"""
+Low Level IMAP Connection
+"""
+
+
+import errno
+import socket
+import subprocess
+import sys
+from io import DEFAULT_BUFFER_SIZE
+from logging import getLogger
+from . import exceptions
+
+try:
+    import ssl
+    HAVE_SSL = True
+except ImportError:
+    HAVE_SSL = False
+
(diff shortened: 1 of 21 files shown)


mjs:
Thanks for this. It looks promising.

I've had a crazy week so I haven't been able to look at the changes properly yet but I will soon.

Issue #390, opened 2019-10-28
Bug returning folder names when they can be considered 'numerical'


baaym:
Hi all,

Given a folder name that could be considered a number, ImapClient will mangle that folder name in the case of leading zeroes.
For example: if I create a folder `012345`, and execute a `list_folders`, ImapClient will return `12345`.

The cause can be traced down to the `atom` function in `response_parser.py`. That function will check if a token in the response is a digit, and then cast the token to an `int`. When the `int` is later cast back to a `str` (in `_proc_folder_list` in `imapclient.py`), the leading zero has already been removed.

I'd be happy to provide a pull request for this bug, but since the `atom` function is a general-purpose function, I'm not sure what the best course of action is. Changing that behavior will likely have side effects on other parts of the code that do expect `int`s being returned.

mjs:
Ouch, good catch! I bet you could also see incorrect behaviour if a folder was called `NIL` :)

I think the right approach is to have a specialised function for handling folder list responses which doesn't use `atom` and just assumes everything is strings. It could still use TokenSource to generate tokens. 

Does that sound reasonable? Are you prepared to have a go at the change?

PR #405, opened 2020-06-17
Prevent converting numbers with leading zeroes to int. (#390)

jap:
This is a minimal attempt to fix bug #390. This fix assumes that when an imap server starts a number with a leading zero, this token should be treated as a string.

As for the suggestion to have a list of attributes that should always be strings:  in our case, we are getting back folder names because we query dovecot for the 'X-MAILBOX' attribute, which is non-standard. I'm not sure if imapclient should maintain/contain a list of (non-standard) attributes that should always be considered a string.

The NIL case might still breaks things, but some manual testing indicates dovecot is smart enough to quote it in that case. I'm not sure if the folder list not being quoted is a dovecot bug or not, but this fix seems easy and makes sense to me.

### DIFF
--- a/imapclient/response_parser.py
+++ b/imapclient/response_parser.py
@@ -215,7 +215,7 @@ def atom(src, token):
         return literal_text
     elif len(token) >= 2 and (token[:1] == token[-1:] == b'"'):
         return token[1:-1]
-    elif token.isdigit() and token[:1] != b'0':
+    elif token.isdigit() and (token[:1] != b'0' or len(token) == 1):
         # this prevents converting items like 0123 to 123
         return int(token)
     else:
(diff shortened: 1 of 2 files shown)


jap:
... ok this was not yet optimal, hold on.

mjs:
Thanks for this. I'll wait for further updates

jap:
Oh sorry there's no need to wait - that comment was my reaction on it failing when a 0 was returned in IMAP and an int was expected on the Python side - with the second commit it is in a good shape (and we've been running this in production at StartMail for over a week now without complaints from Sentry).

mjs:
This obviously isn't a complete fix but it is an improvement. Thanks.

PR #411, opened 2020-09-27
Allow any custom SASL mechanism to be provided

grawity:
This allows mechanisms such as EXTERNAL, GSSAPI, or SCRAM-SHA-256 to be used in the same way as with imaplib.

Tested against Dovecot using SASL GSSAPI (which is a 3-step mechanism).

### DIFF
--- a/imapclient/imapclient.py
+++ b/imapclient/imapclient.py
@@ -386,9 +386,51 @@ class IMAPClient(object):
     def sasl_login(self, mech_name, mech_callable):
         """Authenticate using a provided SASL mechanism (requires server support).
 
-        The 'mech_callable' must accept a parameter with the server challenge
-        and return the client response. It will be called as many times as the
-        server produces challenges.
+        The *mech_callable* will be called with one parameter (the server
+        challenge as bytes) and must return the corresponding client response
+        (as bytes, or as string which will be automatically encoded).
+
+        It will be called as many times as the server produces challenges,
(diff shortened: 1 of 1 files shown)


imapclient/imapclient.py
```diff
@@ -383,6 +383,18 @@ def plain_login(self, identity, password, authorization_identity=None):
         except exceptions.IMAPClientError as e:
             raise exceptions.LoginError(str(e))
 
+    def sasl_login(self, mech_name, mech_callable):
+        """Authenticate using a provided SASL mechanism (requires server support).
+
+        The 'mech_callable' must accept a parameter with the server challenge
+        and return the client response. It will be called as many times as the
+        server produces challenges.
+        """
```
NicolasLM, line 392: Could you add a small usage example in the docstring? It would help to understand how `mech_callable` should be built. Thank you.

NicolasLM:
Perfect, thank you!

grawity:
The GSSAPI client code I used for testing is primarily internal-use and made to work on my own systems, but for reference it can be found here: https://github.com/grawity/code/blob/master/lib/python/nullroute/sec/sasl.py

Other Python SASL modules like https://github.com/thobbs/pure-sasl or https://github.com/dwd/Suelta could also be used for this purpose – e.g. with puresasl.client.SASLClient(), the callable might be `sasl.process` or `lambda ch: sasl.process(ch)`.

PR #422, opened 2020-12-19
Add optional timeout parameter to IMAP4WithTimeout.open

link2xt:
A follow-up to https://github.com/mjs/imapclient/pull/413

PR #421, opened 2020-12-08
Fix IMAP4_TLS for imaplib in Python 3.9+

SpotlightKid:
Trying to use [imapautofiler](https://github.com/imapautofiler/imapautofiler) with python 3.9, I get this error:

```
Traceback (most recent call last):
  File "/home/chris/lib/virtualenvs/imapautofiler/bin/imapautofiler", line 10, in <module>
    sys.exit(main())
  File "/home/chris/work/imapautofiler/imapautofiler/imapautofiler/app.py", line 157, in main
    conn = client.open_connection(cfg)
  File "/home/chris/work/imapautofiler/imapautofiler/imapautofiler/client.py", line 34, in open_connection
    return IMAPClient(cfg)
  File "/home/chris/work/imapautofiler/imapautofiler/imapautofiler/client.py", line 131, in __init__
    self._conn = imapclient.IMAPClient(
  File "/home/chris/lib/virtualenvs/imapautofiler/lib/python3.9/site-packages/imapclient/imapclient.py", line 161, in __init__
    self._imap = self._create_IMAP4()
  File "/home/chris/lib/virtualenvs/imapautofiler/lib/python3.9/site-packages/imapclient/imapclient.py", line 172, in _create_IMAP4
    return tls.IMAP4_TLS(self.host, self.port, self.ssl_context,
  File "/home/chris/lib/virtualenvs/imapautofiler/lib/python3.9/site-packages/imapclient/tls.py", line 171, in __init__
    imaplib.IMAP4.__init__(self, host, port)
  File "/usr/lib/python3.9/imaplib.py", line 202, in __init__
    self.open(host, port, timeout)
TypeError: open() takes 3 positional arguments but 4 were given
```

This is due to a [change in imaplib](https://docs.python.org/3/whatsnew/3.9.html#imaplib) in Python 3.9. The `open()` method now accepts a new optional `timeout` argument and this is also passed when `open()` is called inside `imaplib`. `imapclient.IMAP4_TLS` overwrites `open()` but does not accept a `timeout` argument, optional or not, which leads to the above error.

This PR fixes that by adding the `timeout` argument with a default value of `None` to `open()` and also adds a default value of `None` to the `timeout` argument of the `__init__()` method.

PR #424, opened 2021-01-04
Fix Python 3.9 compatibility

marmarek:
imaplib in Py3.9 adds optional `timeout` argument to `IMAP4.__init__()` and `IMAP4.open()`.
Avoid crashing when it is really used (imaplib calling open internally, with timeout parameter).
Store just connect timeout to make it consistent with `imaplib.IMAP4`.

This is similar change to 28da86caa22b385b1ae5b050cf9aa829ec2eef71

### DIFF
--- a/imapclient/imap4.py
+++ b/imapclient/imap4.py
@@ -9,15 +9,21 @@ import socket
 class IMAP4WithTimeout(imaplib.IMAP4):
 
     def __init__(self, address, port, timeout):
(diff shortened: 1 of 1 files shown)


link2xt:
I opened a similar PR #422 previously.

mjs:
Thanks very much for this but I'm going to take #421 because it addresses the same issue and it came first.

PR #426, opened 2021-01-19
ssl context created with purpose ssl.Purpose.CLIENT_AUTH

pinoatrome:
in case the ssl_context argument is None the create_default_context method should be created with CLIENT_AUTH 'Purpose' argument because the default is SERVER_AUTH

```
    if ssl_context is None:
        ssl_context = ssl.create_default_context(purpose=ssl.Purpose.CLIENT_AUTH)
```


mjs:
Thanks very much.

pinoatrome:
Hi @mjs ,
you welcome.
Thanks to you (and all the contributors) for this great project.

PR #444, opened 2021-06-30
fixed minor typo and added a doc in example

shoaib30:
Added a comment in the examples while going through them. And removed a stray ` in the Readme


mjs:
Thanks very much!

PR #452, opened 2021-11-25
Fix undiscoverable tests due to name shadowing

BoniLindsley:
There were tests with the identical names, but with `POLL_SUPPORT` patched to `False`.

Changing the test names increased the number of discovered tests by three.


NicolasLM:
Again, thank you!

mjs:
Thanks!

PR #458, opened 2021-11-27
Fix render artefacts in changelog

BoniLindsley:
Building documentation was emitting warnings from the release notes page.

```
/home/boni/base/src/imapclient/doc/src/releases.rst:13: WARNING: Unexpected indentation.
/home/boni/base/src/imapclient/doc/src/releases.rst:14: WARNING: Block quote ends without a blank line; unexpected unindent.
```

It seems that sublists must be wrapped by newlines in reStructuredText. Without the newlines, they are treated as text of the preceding bullet point. This can be seen in the hosted documentation: <https://imapclient.readthedocs.io/en/2.2.0/releases.html#changed>. 

The other commit removes leading indentation from two bullet lists. It seems indented lines on their own are automatically blockquotes in reStructuredText.

I took the opportunity to change the lists to use `-` instead of `*`, to fit the text style of the newer changelogs. They happen to be the latest lists to use `*`, so they should fit in fine.

### DIFF
--- a/doc/src/releases.rst
+++ b/doc/src/releases.rst
@@ -356,26 +356,26 @@ greatly speeds up parsing time.
 
 Other
 -----
-  * Perform quoting of Gmail labels. Thanks to Pawel Sz for the fix.
-  * The type of the various flag constants was fixed. Thanks to Thomi
-    Richards for pointing this out.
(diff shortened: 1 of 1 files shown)


NicolasLM:
Again, thank you.

PR #460, opened 2021-12-05
Simplify running tests against all Python versions

mjs:
- Added tox-all script which runs all unit tests against all supported
  Python versions using Docker
- Adjusted tox.ini so that old pip and setuptools versions are used
  where required for old Python versions
- Updated docs to mention tox-all


mjs:
@BoniLindsley

BoniLindsley:
Nice way to run Python 3.4. Hopefully, I will not need it much.

I am not too keen on pulling in Sphinx for testing, but it works.

NicolasLM:
Makes sense, merged.

PR #462, opened 2021-12-30
added flags and internaldate capabilites for MULTIAPPEND

d70-t:
While in #195 it was suggested to implement a `multiappend` function in a way supporting flags and internal date, this was not implemented in #399.

This PR adds the suggested functionality (iterator over `{'msg': ...[, 'flags': ...][, 'date': ...]}`-dicts) in a backwards-compatible manner.

### DIFF
--- a/imapclient/imapclient.py
+++ b/imapclient/imapclient.py
@@ -1424,8 +1424,11 @@ class IMAPClient(object):
     def multiappend(self, folder, msgs):
         """Append messages to *folder* using the MULTIAPPEND feature from :rfc:`3502`.
 
-        *msgs* should be a list of strings containing the full message including
-        headers.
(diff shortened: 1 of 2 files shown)


mjs:
Thanks very much. I’m on holidays at the moment with limited connectivity but will take a proper look on my return.

mjs (changes requested):
This looks great. Thanks for adding tests as well.

 Would you mind updating the docstring to mention the approach of passing a dict with the internaldate and flags? Once that's in place this is good to merge.

d70-t:
Thanks for reviewing @mjs. I've made the requested changes to the docstring.

mjs (approved):
Perfect, thanks!

PR #477, opened 2022-06-23
chore: setup `black` for CI

JohnVillalovos:
Commit f0723402d00c5612a428d3815e4f1e653eb68ff0 ran `black` on the
code base. But since it wasn't being enforced in the CI, it has
drifted away from `black` standarized formatting.

Did the following:
  * Added a Github CI job to run `black`
  * Added a `black` tox environment
  * Ran `black` on the code
  * Added the `requirements-dev.txt` file

This has been setup so that it should be easy to add other
checkers/linters in the future as desired.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,30 @@
+name: Lint
+
+# If a pull-request is pushed then cancel all previously running jobs related
+# to that pull-request
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
+  cancel-in-progress: true 
+
+on:
+  push:
+    branches:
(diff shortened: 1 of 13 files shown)


NicolasLM:
Could we put all the dependencies in a `requirements-dev.txt` just to make it clear that the project doesn't have any dependency when using it as a library?

mjs (commented):
This looks great aside from the concern Nicolas raised.
Thanks very much!

JohnVillalovos:
> Could we put all the dependencies in a `requirements-dev.txt` just to make it clear that the project doesn't have any dependency when using it as a library?

Done! 😊

NicolasLM:
Perfect, thanks!

PR #487, opened 2022-07-21
Remove configparser deprecation warnings

### LINKED ISSUE #408: configparser related deprecation messages
https://github.com/mjs/imapclient/blob/5b1f5fd35fc9ad7d87fd5c059023335fadf4377e/imapclient/config.py#L49

```
>>> from imapclient.config import *
>>> parse_config_file("/tmp/foo.ini")
/root/checked_repos/imapclient/imapclient/config.py:47: DeprecationWarning: The SafeConfigParser class has been renamed to ConfigParser in Python 3.2. This alias will be removed in future versions. Use ConfigParser directly instead.
  parser = SafeConfigParser(get_string_config_defaults())
/root/checked_repos/imapclient/imapclient/config.py:49: DeprecationWarning: This method will be removed in future versions.  Use 'parser.read_file()' instead.
```
Ref : https://docs.python.org/3/library/configparser.html#configparser.ConfigParser.readfp

claudep:
Fixes #408

### DIFF
--- a/imapclient/config.py
+++ b/imapclient/config.py
@@ -42,9 +42,8 @@ def parse_config_file(filename):
     Used by livetest.py and interact.py
     """
 
-    parser = configparser.SafeConfigParser(get_string_config_defaults())
-    with open(filename, "r") as fh:
-        parser.readfp(fh)
+    parser = configparser.ConfigParser(get_string_config_defaults())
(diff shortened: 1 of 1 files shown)


mjs:
Thanks, this is great!

PR #492, opened 2022-09-27
chore: convert from `optparse` to `argparse`

JohnVillalovos:
The `optparse` library has been deprecated since Python 3.2 [1]

Convert `optparse` usage to the `argparse` library.

[1] https://docs.python.org/3/library/optparse.html


mjs (approved):
Nice. Thank you!

PR #508, opened 2023-08-15
New function `uid_expunge`, which requires the capability UIDPLUS.

### LINKED ISSUE #496: expunge func assumes UIDPLUS capability and uses UID EXPUNGE in error
Hi guys,
Firstly, thanks for a fine piece of code that's extremely vital to the community. You are to be commended and I'm incredibly grateful.

However, I think I have detected a small error that has some ramifications which break certain mass-move/delete operations on IMAP4 servers that do not yet support UIDPLUS (https://github.com/hmailserver/hmailserver being one such)

It appears that the imapclient.expunge function here: https://github.com/mjs/imapclient/blob/d0f5c4e927ddda1972066ef04242a09d2dc2a666/imapclient/imapclient.py#L1469
...assumes that if it is given a list of UIDs by a caller, then it is safe to just go ahead and use UID EXPUNGE command-methodology to work through the list and expunge them, without actually confirming that UIDPLUS (and thereby UID EXPUNGE) is even available on the current IMAP server.

This is potentially unsafe, though thankfully not destructive, but it will derail operations that are working through a mailbox folder, expecting batches of messages to have been moved/deleted, such as DMARC report processing. See [parsedmarc](https://domainaware.github.io/parsedmarc/) as an example which relies, ultimately, on IMAPClient, and (imho acceptably) assumes that a UID list given to an interim mail-handler ([Mailsuite](https://seanthegeek.github.io/mailsuite/) will be taken care of, in terms of dealing with any UIDPLUS capability or not). Mailsuite evidently makes the same assumption, and defers the problem to IMAPClient, which makes sense, really, as IMAPClient is 'nearer the metal'.

As an aside, the RFCs referred to in IMAPClient's expunge function have been obsoleted::
https://www.rfc-editor.org/rfc/rfc9051#section-6.4.3 now defines the EXPUNGE command (old style global expunge) in the current IMAP RFC
https://www.rfc-editor.org/rfc/rfc9051#section-7.5.1 describes responses from IMAP servers when using EXPUNGE

Not a lot has changed (in terms of EXPUNGE), but you might like to point to the newer RFC anyway.

https://www.rfc-editor.org/rfc/rfc4315#section-2.1 describes the newer UID EXPUNGE command, which is part of UIDPLUS and not a de facto capability of all IMAP servers yet, sadly. Most usefully, however, it gives an alternative for handling lists of message UIDs, in the event that UID EXPUNGE isn't available:

```
      If the server does not support the UIDPLUS capability, the client
      should fall back to using the STORE command to temporarily remove
      the \Deleted flag from messages it does not want to remove, then
      issuing the EXPUNGE command.  Finally, the client should use the
      STORE command to restore the \Deleted flag on the messages in
      which it was temporarily removed.
```

**Would it be possible to beg that IMAPClient is updated to fall back to that method in the event that UIDPLUS capability is not found on the current IMAP server, please?** 

Older non-UIDPLUS-capable IMAP servers like hmailserver will misidentify the use of a UID EXPUNGE command as a broken invocation of the UID command, and report a bad or missing uid number, and not expunge anything at all. It parses the word 'UID', expects a number, then parses 'EXPUNGE' and assumes failure, and dismisses the range of UIDs that follows. 

Net result is that the messages concerned are not purged from the mailbox folder, and will remain there being detected by batching operations that are trying to process the whole folder in chunks, saving processed-output in stages - for example, a pile of DMARC reports. These tools will end up iterating the same set of emails over and over again until something else on the IMAP server randomly happens to call a global EXPUNGE (old RFC9051 method), at which point, they'll vanish and the DMARC analysis process will continue - but only for the next batch, then it's back to looping again!

If you are able to fix this, it would be greatly appreciated by a number of users, I'm sure (and definitely by me!) Processing of DMARC reports for smaller business and homeworkers who use their own open-source mailservers on Windows may be a small subset of global users, but frankly anything that makes life easier for the detection and prevention of spam and spoofed emails, has to be a good thing for the community at large, no?

Of course it's to be hoped that hmailserver eventually gains UIDPLUS functionality, but at the current time, it's not evident, and in any case, there may be other IMAP servers out there in the wild that don't. Supporting of UIDPLUS is still optional; whereas the dependence upon it without checking its availability is technically 'broken behaviour'. So hopefully IMAPClient can smooth the way with a fix?

Thanks for reading! If you need any help testing, please get in touch and I'll try and assist.
Best regards
TBJ

axoroll7:
The function `expunge` can raise an error at the server side, when the function is utilized with *messages*, and the server does not support the capability UIDPLUS.

Closes #496

### DIFF
--- a/imapclient/imapclient.py
+++ b/imapclient/imapclient.py
@@ -1507,15 +1507,14 @@ class IMAPClient(object):
 
     @require_capability("UIDPLUS")
     def uid_expunge(self, messages):
-        """Same functionality as ``expunge``,
-        but *messages* must be specified,
-        and the capability UIDPLUS is tested beforehand.
-        
+        """Same functionality as ``expunge``, but *messages* must be
+        specified, and the capability UIDPLUS is tested beforehand. It should
+        be more fail-proof than ``expunge`` with *messages*, which cannot be
+        updated to prevent breaking compability with existing codebase.
+
         See :rfc:`4315#section-2.1` section 2.1 for more details.
         """
-        return self._command_and_check(
-            "EXPUNGE", join_message_ids(messages), uid=True
-        )
+        return self._command_and_check("EXPUNGE", join_message_ids(messages), uid=True)
 
     @require_capability("ACL")
     def getacl(self, folder):
--- a/imapclient/imapclient.py
+++ b/imapclient/imapclient.py
@@ -1505,6 +1505,18 @@ class IMAPClient(object):
         tag = self._imap._command("EXPUNGE")
         return self._consume_until_tagged_response(tag, "EXPUNGE")
 
+    @require_capability("UIDPLUS")
+    def uid_expunge(self, messages):
+        """Same functionality as ``expunge``,
+        but *messages* must be specified,
+        and the capability UIDPLUS is tested beforehand.
+        
+        See :rfc:`4315#section-2.1` section 2.1 for more details.
+        """
+        return self._command_and_check(
+            "EXPUNGE", join_message_ids(messages), uid=True
+        )
+
     @require_capability("ACL")
     def getacl(self, folder):
         """Returns a list of ``(who, acl)`` tuples describing the
--- a/imapclient/imapclient.py
+++ b/imapclient/imapclient.py
@@ -1467,7 +1467,10 @@ class IMAPClient(object):
         )
 
     def expunge(self, messages=None):
-        """When, no *messages* are specified, remove all messages
+        """Use of the *messages* argument is discouraged.
+        Please see the ``uid_expunge`` method instead.
+
+        When, no *messages* are specified, remove all messages
         from the currently selected folder that have the
         ``\\Deleted`` flag set.
 
@@ -1475,9 +1478,9 @@ class IMAPClient(object):
         followed by a list of expunge responses. For example::
 
             ('Expunge completed.',
-             [(2, 'EXPUNGE'),
-              (1, 'EXPUNGE'),
-              (0, 'RECENT')])
+            [(2, 'EXPUNGE'),
+            (1, 'EXPUNGE'),
+            (0, 'RECENT')])
 
         In this case, the responses indicate that the message with
         sequence numbers 2 and 1 where deleted, leaving no recent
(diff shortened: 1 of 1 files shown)


mjs:
Thanks very much for this. The change looks good but could you address the formatting issue mentioned by black?

mjs:
Also, I've just fixed up the project's Github Actions config so if you rebase onto master or merge master the actions should work now.

axoroll7:
Hello,
Thank you for your amazing library !

I updated the pull request, to address the linter issue.
I added some context in the function description.
English is not my first language, I would like you to read carefully each word to correct me.

imapclient/imapclient.py
```diff
@@ -1505,6 +1505,17 @@ def expunge(self, messages=None):
         tag = self._imap._command("EXPUNGE")
         return self._consume_until_tagged_response(tag, "EXPUNGE")
 
+    @require_capability("UIDPLUS")
+    def uid_expunge(self, messages):
+        """Same functionality as ``expunge``, but *messages* must be
+        specified, and the capability UIDPLUS is tested beforehand. It should
+        be more fail-proof than ``expunge`` with *messages*, which cannot be
+        updated to prevent breaking compability with existing codebase.
```
mjs, line 1513: I don't think the docstring requires quite as much detail. Something like this should do:

>  Expunge deleted messages with the specified message ids from the folder. This requires the UIDPLUS capability.
>
> See :rfc:`4315#section-2.1` section 2.1 for more details.

 Then please extend the docstring on the expunge method to discourage use of `messages` there. Something like:

> Use of the *messages* argument is discouraged. Please see the uid_expunge method instead.

I would be ok with this replacing the existing paragraph about the `messages` argument.

Thanks!

axoroll7: @mjs 
Your choice of words is better than mine, I am using it without modification. New commit e0cbf56 and ea5ef55

axoroll7:
@mjs 
Your choice of words is better than mine, I am using it without modification. New commit e0cbf56 and ea5ef55

axoroll7:
Depreciation notice, first line ?

mjs (approved):
This is great. Thanks very much.

PR #516, opened 2023-08-19
chore: add `dependabot.yml` to enable automatic PRs

JohnVillalovos:
Add a basic `dependabot.yml` so that PRs will be automatically generated to update the versions of the dependencies.

More information here:
https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuring-dependabot-version-updates

Once this is merged then Dependabot will propose PRs to update
dependencies.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "weekly"


mjs:
Thanks, although I 'm not sure  how useful this is. IMAPClient's only external dependency is Sphinx for building the docs. 

Do we really need dependabot?

JohnVillalovos:
There is black, flake8, isort, and in the future mypy

JohnVillalovos:
At the moment black and flake8 can be updated to newer versions. In my testing when this got merged into my fork Dependabot created a PR to update black and flake8.

NicolasLM:
I also don't see much value in that and I believe it will generate a fair amount of noise. Maybe let's not pin development dependencies?

JohnVillalovos:
> I also don't see much value in that and I believe it will generate a fair amount of noise. Maybe let's not pin development dependencies?

It can be setup to only propose PRs once a day, week, month. So the noise can be quite minimal. I have not configured that in this PR.

The problem with not pinning is that things break unexpectedly. One day things are working. Another day someone proposes a PR and it fails for them because a dependency has changed and it breaks the CI. Like right now when black gets updated it will break CI as there was a minor change.

JohnVillalovos:
Updated the PR to have Dependabot run once a week.

mjs:
I'm happy for this to run once a week. 

One thing though: will dependabot find the dependencies? From my quick read of the docs, it looks to me that the pip package-ecosystem checks look in requirement.txt and pyproject.toml. IMAPClient only has the dev dependencies in requirements-dev.txt.

Maybe we should move the dependencies to pyproject.toml as optional dependencies as described in PEP621. Apparently pip can read them from there (I haven't tried it).

Really, we should be doing away with setup.py in favour of pyproject.toml but that's perhaps a separate project.

JohnVillalovos:
> One thing though: will dependabot find the dependencies? From my quick read of the docs, it looks to me that the pip package-ecosystem checks look in requirement.txt and pyproject.toml. IMAPClient only has the dev dependencies in requirements-dev.txt.

Yes it finds them there. When I merged it into my fork it created these two PRs in my fork. I'm guessing it looks for `requirements*.txt` or something like that.

 * https://github.com/JohnVillalovos/imapclient/pull/3
 * https://github.com/JohnVillalovos/imapclient/pull/4

> Maybe we should move the dependencies to pyproject.toml as optional dependencies as described in PEP621. Apparently pip can read them from there (I haven't tried it).
> 
> Really, we should be doing away with setup.py in favour of pyproject.toml but that's perhaps a separate project.

Both of those sound like probably good ideas for the future.

mjs:
Thanks very much.  :star:

PR #518, opened 2023-08-20
Initial framework for adding type-hints and added type-hints to `version.py`

JohnVillalovos:
This is adding the initial framework for adding type-hints. It adds the running of `mypy` and type-checks nothing in the first commit.

Then adds type-checking of the file `imapclient/version.py` and others can use that as an example of how to add type-hints by updating `pyproject.toml`

### DIFF
--- a/imapclient/version.py
+++ b/imapclient/version.py
@@ -2,10 +2,12 @@
 # Released subject to the New BSD License
 # Please see http://en.wikipedia.org/wiki/BSD_licenses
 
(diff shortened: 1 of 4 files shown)


NicolasLM:
Looks good to me, although given the low activity on IMAPClient in the past few of years, I doubt type hints will ever be populated on a significant portion of the code.

mjs:
Still worth it I think

PR #530, opened 2023-08-27
chore: add type-hints to `response_*.py`

JohnVillalovos:
Added type-hints to:
  * `response_lexer.py`
  * `response_parser.py`
  * `response_types.py`


JohnVillalovos:
This took quite a long time to do. So possible a type-hint is incorrect. But usually we will catch those as we complete all the type-hints. I think everything is correct, but no guarantees 🙂

mjs (approved):
This looks great. As you said, we'll catch any typing errors as more of the code gets hints.

PR #531, opened 2023-08-27
chore: rename type-var to lessen possible confusion

JohnVillalovos:
Rename the type-var in `utils.py` from `_Atom` to `_TupleAtom` so it isn't confused with the `_Atom` in `typing_imapclient.py`


mjs (approved):
I agree that's clearer. Great!

PR #532, opened 2023-08-27
chore: remove `pop_with_default()` function

JohnVillalovos:
In current Pythons we can use `dict.pop(key, default)`

Also converted some `**kwargs` to two named arguments with defaults.

### DIFF
--- a/imapclient/imapclient.py
+++ b/imapclient/imapclient.py
@@ -1572,7 +1572,7 @@ class IMAPClient:
         quota_root_rep = self._raw_command_untagged(
             b"GETQUOTAROOT", to_bytes(mailbox), uid=False, response_name="QUOTAROOT"
         )
-        quota_rep = pop_with_default(self._imap.untagged_responses, "QUOTA", [])
+        quota_rep = self._imap.untagged_responses.pop("QUOTA", [])
         quota_root_rep = parse_response(quota_root_rep)
         quota_root = MailboxQuotaRoots(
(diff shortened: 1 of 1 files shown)


axoroll7:
Which current Pythons ?

JohnVillalovos:
> Which current Pythons ? Edit : Works with 3.6

All the ones that are supported by the current version of `imapclient`

axoroll7:
> > Which current Pythons ? Edit : Works with 3.6
> 
> All the ones that are supported by the current version of `imapclient`

Ok thanks, I should have known

JohnVillalovos:
As a bit of trivia. It was added in Python 2.3.

https://docs.python.org/2.7/library/stdtypes.html#mapping-types-dict

imapclient/imapclient.py
```diff
@@ -1721,11 +1721,9 @@ def _send_literal(self, tag, item):
         logger.debug("   (literal) > %s", debug_trunc(item, 256))
         self._imap.send(item)
 
-    def _command_and_check(self, command, *args, **kwargs):
-        unpack = pop_with_default(kwargs, "unpack", False)
-        uid = pop_with_default(kwargs, "uid", False)
-        assert not kwargs, "unexpected keyword args: " + ", ".join(kwargs)
```
mjs, line 1727: I think at some point kwargs was passed down to other calls but not any more so what you've done makes much more sense.

mjs (approved):
A nice little clean up. Thanks.

Issue #517, opened 2023-08-19
Special folder attributes, even when SPECIAL-USE is not defined


axoroll7:
In the source code, imapclient check the capability SPECIAL-USE before searching special folder attributes from the output of the command LIST **non-extended**. But the capability SPECIAL-USE should not be tested when searching special attributes from the **non-extended** LIST.

While reading the RFC 6154, I come across a line that said the SPECIAL-USE capability is mandatory with the **extended** LIST command, but not with the **non-extended** LIST command :
> RFC 6154, chapter 2.0, in the first paragraph :
> There is no capability string related to the support of special-use attributes on the non-extended LIST command.

With the email provider *Zoho mail*, folders have special folder attributes, with the output of the command LIST **non-extended**, but the capability SPECIAL-USE is not present. *Zoho mail* has a important marked share (maybe 1.58% ?) : [Zoho mail market share](https://www.google.com/search?q=zoho+mail+market+share)

When the **extended** LIST will be implemented, the SPECIAL-USE should should be tested, but only with the **extended** LIST.

NicolasLM:
Hello, you seem to be correct. Feel free to submit a PR for that.

axoroll7:
pull request #533

PR #533, opened 2023-08-27
Removed SPECIAL-USE capability check

axoroll7:
To comply with RFC 6154. Only the extended LIST command need SPECIAL-USE, not the non-extended LIST command (section 2.0, first paragraph).


axoroll7:
issue #517

mjs (approved):
Thank you!

PR #534, opened 2023-08-28
chore: convert `namedtuple` to `dataclass`

JohnVillalovos:
A `dataclass` is nicer than a `namedtuple`. Makes it is easy to add type-hints and also they have the added flexibility to customize things if desired.

### DIFF
--- a/imapclient/imapclient.py
+++ b/imapclient/imapclient.py
@@ -2,6 +2,7 @@
 # Released subject to the New BSD License
 # Please see http://en.wikipedia.org/wiki/BSD_licenses
 
+import dataclasses
 import functools
 import imaplib
 import itertools
@@ -10,11 +11,10 @@ import select
 import socket
 import sys
 import warnings
-from collections import namedtuple
 from datetime import date, datetime
 from logging import getLogger, LoggerAdapter
 from operator import itemgetter
-from typing import Optional
+from typing import List, Optional
 
 from . import exceptions, imap4, response_lexer, tls
 from .datetime_util import datetime_to_INTERNALDATE, format_criteria_date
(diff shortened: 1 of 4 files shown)


JohnVillalovos:
The one possible issue I see with this is if people are using these as `tuples` and using indexing instead dot attributes.

mjs:
> The one possible issue I see with this is if people are using these as `tuples` and using indexing (example `variable[1]`) instead of dot attributes (example `variable.attribute`).
> 
> A work around for that would be to add a method like this. But not sure...
> 
> ```
> def __getitem__(self, index):
>     return dataclasses.astuple(self)[index]
> ```
> 
> If we did that I would think to add a deprecation warning to tell people to use attribute access.

I wouldn't bother with this because:
1. I very much doubt that anyone is accessing attributes positionally
1. The major version number will be incremented for the next release so we're allowed to break compatibility :)

Let's not clutter things up if we don't need to.

mjs:
I'm ok with this PR as it is although it has made me notice all the `if TYPE_CHECKING` parts which make the code harder to read. Do we really need to leave those in?

mjs:
I should have also said that I like the overall change to dataclass. It's much more readable.

JohnVillalovos:
> I'm ok with this PR as it is although it has made me notice all the `if TYPE_CHECKING` parts which make the code harder to read. Do we really need to leave those in?

I cleaned those up just now to make less of them. But they are needed unless want the `assert()` calls to run in live code.

JohnVillalovos:
@mjs Let me know if there is anything you want me to change. Thanks.

mjs:
Thanks for this. Merging...

mjs (approved):

PR #537, opened 2023-09-10
Add type-hints to `setup.py` and `imapclient/config.py`. Also use `argparse.Namespace` instead of `Bunch`

JohnVillalovos:


JohnVillalovos:
@mjs / @NicolasLM Would be interested to hear if you use a type-aware editor and if you are noticing any more assistance with the type-hints we have added.

I'm saving `imapclient.py` for last as it has 290 reported issues when I enable checking of it 😄 But we only have a total of four files to do after this PR.

mjs (approved):
Fantastic

mjs:
> @mjs / @NicolasLM Would be interested to hear if you use a type-aware editor and if you are noticing any more assistance with the type-hints we have added.

I haven't noticed much benefit yet because I've mainly been in imapclient.py, but thank you anyway.

PR #586, opened 2024-06-21
Move man page to section 7

wiene:
The Debian lint tool `lintian` noticed that `imapclient` puts its man page to section 1 although this section is intended for
> Executable programs or shell commands

to quote `man man`.

This patch moves the man page to section 7 which seems more suitable.


mjs (approved):
That's great, thanks!

PR #629, opened 2025-12-06
Simplify IMAP4_TLS class and fix Python 3.14+ compatibility

### LINKED ISSUE #618: AttributeError: property 'file' of 'IMAP4_TLS' object has no setter on Python 3.14
Running this utility in the scope of some other package (`parsedmarc`), I ran into an issue of it failing with the following message:
`AttributeError: property 'file' of 'IMAP4_TLS' object has no setter`.

From what I could quickly gather, this is triggered by the following call:
https://github.com/mjs/imapclient/blob/391cc6d66d35c16bed11292ff00835646cd50ae5/imapclient/tls.py#L56

This call no longer seems to be valid in Python 3.14 due to changes in the `imaplib` library to the `file` property:
https://github.com/python/cpython/blob/5cea8435943b4f9d22c89c80d86c2ba3b392c6f9/Lib/imaplib.py#L322-L336
(see also https://github.com/python/cpython/commit/0fef47e5bbd167c21eb4f3cbd885cf61270014e7).

seanthegeek:
Fix #618 without needing to access undocumented Python `imaplib` attributes that should not be modified, such as `file`. Tested against Gmail IMAP TLS on Python 3.8, 3.13, and 3.14 using this test script

```python
import ssl
from imapclient import IMAPClient


HOST = "imap.gmail.com"
PORT = 993


def main() -> None:
    ctx = ssl.create_default_context()
    with IMAPClient(
        HOST,
        port=PORT,
        ssl=True,
        ssl_context=ctx,
        timeout=30,
    ) as client:
        print("Connected to:", HOST, PORT)
        print("welcome:", client.welcome)

        caps = client.capabilities()
        print("capabilities:", caps)
        print("Successfully connected and ran CAPABILITY without auth.")


if __name__ == "__main__":
    main()
```

## See also

https://github.com/python/cpython/issues/142307
https://github.com/seanthegeek/mailsuite/issues/15
https://github.com/domainaware/parsedmarc/issues/636


mjs:
Thanks! There's still a flake8 issue (unused import) but I'll merge and fix this myself.

mjs:
Release 3.1.0 has just been done and includes this fix. Thanks!

Issue #564, opened 2023-11-19
KeyError: b'RFC822'


sidamos:
This code threw an unexpected exception:
                    messages = fetchServer.search("UNSEEN")
                    for uid, message_data in fetchServer.fetch(messages, "RFC822").items():
                        email_message = email.message_from_bytes(message_data[b"RFC822"])
->

Traceback (most recent call last):
  File "/work2/imap-notify/./imap-notify.py", line 68, in <module>
    email_message = email.message_from_bytes(message_data[b"RFC822"])
                                             ~~~~~~~~~~~~^^^^^^^^^^^
KeyError: b'RFC822'

I cannot reproduce it with the same mails.

mjs:
This is likely the same as #334

PR #647, opened 2026-08-28
fix: handle unsolicited responses in fetch

### LINKED ISSUE #334: Fetch return result unrelated to fetch command params.
Hello, i'm using imapclient 1.1 (last stable, don't know when 2.0 will be released in pip) .
I discovered a weird bug when using : 

```
for msgid, data in imapc.fetch(uids, ['BODY.PEEK[]']).items():
    msg = message_from_bytes(data[b'BODY[]'])
```

sometime data return KeyError due to lack of b'BODY'. So i checked and i found a `data` object related to what i'm doing with an other IMAP client at the same time (modify flags of messages).

This behaviour is really weird. It should be either explained in doc or results should be correctly filtered to avoid this type of problem.

raboof:
Filter out unsolicited responses for messages other than the ones being queried.

Unsolicited responses for messages that are being queried are already handled correctly (the test `test_same_message_appearing_multiple_times` tests that they are correctly merged)

This fixes methods like `get_flags` and `get_gmail_labels` that, when `fetch` returned unsolicited data, would crash (when the unsolicited data did not contain the queried fields), or in theory even return incorrect data (when the unsolicited data would contain the queried fields for other messages - though that sounds unlikely in practice).

Fixes #334 and possibly #564

### DIFF
--- a/imapclient/util.py
+++ b/imapclient/util.py
@@ -3,7 +3,7 @@
 # Please see http://en.wikipedia.org/wiki/BSD_licenses
 
 import logging
-from typing import Iterator, Optional, Tuple, Union
+from typing import Iterator, Optional, Sequence, Tuple, Union
 
 from . import exceptions
 
@@ -22,6 +22,19 @@ def to_unicode(s: Union[bytes, str]) -> str:
             )
             return s.decode("ascii", "ignore")
     return s
+
+
+def to_ints(
+    messages: Union[bytes, str, int, Sequence[Union[bytes, str, int]]],
+) -> Sequence[int]:
+    """Convert a sequence of values (such as messages)
+    or a single integer value into an sequence of ints
+    """
+    if not messages:
+        return []
+    if isinstance(messages, (int, str, bytes)):
+        return [int(messages)]
+    return [int(i) for i in messages]
 
 
 def to_bytes(s: Union[bytes, str], charset: str = "ascii") -> bytes:
--- a/imapclient/imapclient.py
+++ b/imapclient/imapclient.py
@@ -22,7 +22,7 @@ from .datetime_util import datetime_to_INTERNALDATE, format_criteria_date
 from .imap_utf7 import decode as decode_utf7
 from .imap_utf7 import encode as encode_utf7
 from .response_parser import parse_fetch_response, parse_message_list, parse_response
-from .util import assert_imap_protocol, chunk, to_bytes, to_unicode
+from .util import assert_imap_protocol, chunk, to_bytes, to_ints, to_unicode
 
 if hasattr(select, "poll"):
     POLL_SUPPORT = True
@@ -1375,6 +1375,9 @@ class IMAPClient:
                     b'INTERNALDATE': datetime.datetime(2011, 2, 24, 19, 30, 36),
                     b'SEQ': 110}}
 
+        Note that due to IMAP unsolicited FETCH responses the return
+        value may include fields not matching the original data query.
+        It is up to the caller to handle that gracefully.
         """
         if not messages:
             return {}
@@ -1391,7 +1394,13 @@ class IMAPClient:
         typ, data = self._imap._command_complete("FETCH", tag)
         self._checkok("fetch", typ, data)
         typ, data = self._imap._untagged_response(typ, data, "FETCH")
-        return parse_fetch_response(data, self.normalise_times, self.use_uid)
+        response = parse_fetch_response(data, self.normalise_times, self.use_uid)
+        # Drop unsolicited responses for other messages
+        return {
+            message_id: response[message_id]
+            for message_id in to_ints(messages)
+            if message_id in response
+        }
 
     def append(self, folder, msg, flags=(), msg_time=None):
         """Append a message to *folder*.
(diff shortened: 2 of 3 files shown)


raboof:
Moving to 'draft' because I unexpectedly may have seen another instance of the issue in get_gmail_labels even with this change, investigating.

raboof:
(found and fixed another instance of the same issue)

raboof:
Encountered a case where in `inbox.get_gmail_labels([single_uid])`, `{ message_id: response[message_id] for message_id in messages }` failed because that uid wasn't in the `response`. Shouldn't it return an element for that message id with an empty list, instead? Is this a sign it may have missed a response from the IMAP server?

raboof:
observed:

```
2026-09-01 08:52:02,225 > b'JABA168 UID FETCH 148622 (BODY.PEEK[])'
2026-09-01 08:52:03,054 < b'* 88 EXISTS'
2026-09-01 08:52:03,054         matched b'\\* (?P<data>\\d+) (?P<type>[A-Z-]+)( (?P<data2>.*))?' => (b'88', b'EXISTS', None, None)
2026-09-01 08:52:03,054 untagged_responses[EXISTS] 38 += ["b'88'"]
2026-09-01 08:52:03,054 < b'* 39 FETCH (UID 148596 FLAGS (\\Seen))'
2026-09-01 08:52:03,054         matched b'\\* (?P<data>\\d+) (?P<type>[A-Z-]+)( (?P<data2>.*))?' => (b'39', b'FETCH', b' (UID 148596 FLAGS (\\Seen))', b'(UID 148596 FLAGS (\\Seen))')
2026-09-01 08:52:03,054 untagged_responses[FETCH] 0 += ["b'39 (UID 148596 FLAGS (\\Seen))'"]
2026-09-01 08:52:03,054 < b'* 46 FETCH (UID 148606 FLAGS (\\Seen))'
2026-09-01 08:52:03,054         matched b'\\* (?P<data>\\d+) (?P<type>[A-Z-]+)( (?P<data2>.*))?' => (b'46', b'FETCH', b' (UID 148606 FLAGS (\\Seen))', b'(UID 148606 FLAGS (\\Seen))')
2026-09-01 08:52:03,054 untagged_responses[FETCH] 1 += ["b'46 (UID 148606 FLAGS (\\Seen))'"]
2026-09-01 08:52:03,054 < b'* 84 FETCH (UID 148759 FLAGS (\\Seen))'
2026-09-01 08:52:03,054         matched b'\\* (?P<data>\\d+) (?P<type>[A-Z-]+)( (?P<data2>.*))?' => (b'84', b'FETCH', b' (UID 148759 FLAGS (\\Seen))', b'(UID 148759 FLAGS (\\Seen))')
2026-09-01 08:52:03,054 untagged_responses[FETCH] 2 += ["b'84 (UID 148759 FLAGS (\\Seen))'"]
2026-09-01 08:52:03,054 < b'* 85 FETCH (UID 148760 FLAGS (\\Seen))'
2026-09-01 08:52:03,055         matched b'\\* (?P<data>\\d+) (?P<type>[A-Z-]+)( (?P<data2>.*))?' => (b'85', b'FETCH', b' (UID 148760 FLAGS (\\Seen))', b'(UID 148760 FLAGS (\\Seen))')
2026-09-01 08:52:03,055 untagged_responses[FETCH] 3 += ["b'85 (UID 148760 FLAGS (\\Seen))'"]
2026-09-01 08:52:03,055 < b'JABA168 OK Success'
2026-09-01 08:52:03,055         matched b'(?P<tag>JABA\\d+) (?P<type>[A-Z]+) (?P<data>.*)' => (b'JABA168', b'OK', b'Success')
```

I guess that can happen then, and we need to account for the possibility that message id's may be missing from the response.

imapclient/util.py
```diff
@@ -24,6 +24,29 @@ def to_unicode(s: Union[bytes, str]) -> str:
     return s
 
 
+def to_int(message: Union[bytes, str, int]) -> int:
+    if isinstance(message, int):
+        return message
+    elif isinstance(message, str):
+        return int(message)
+    elif isinstance(message, bytes):
+        return int.from_bytes(message, byteorder="big", signed=False)
```
mjs: I'm pretty sure this isn't correct. A bytes representation of a message id here would be something `b'42'` not a raw integer. 

```suggestion
    elif isinstance(message, (str, bytes)):
        return int(message)
```

and what about an else clause which raises a ValueError in case an unexpected type is encountered. That would be better than returning `None` as happens now.

imapclient/util.py
```diff
@@ -24,6 +24,29 @@ def to_unicode(s: Union[bytes, str]) -> str:
...
+def to_ints(
+    messages: Union[bytes, str, int, Sequence[Union[bytes, str, int]]],
+) -> Sequence[int]:
+    """Convert a sequence of values (such as messages)
+    or a single integer value into an sequence of ints
+    """
+    if not messages:
+        return []
+    elif isinstance(messages, (int, str, bytes)):
+        return (to_int(messages),)
```
mjs: Best to always return a list instead of sometimes returning a tuple.


```suggestion
        return [to_int(messages)]
```

mjs (commented):
Thanks for tackling this! 

I've highlighted a few minor issues which need resolving but there isn't much required to get this merged. Also note the import ordering lint failure.

raboof:
Thanks for having a look! Updated!

mjs (approved):
Thanks!!

PR #658, opened 2026-09-10
Reject NUL, CR and LF in _raw_command arguments before sending

### LINKED ISSUE #657: search criteria containing CR LF go on the wire inline and become extra IMAP commands
`_normalise_search_criteria` (imapclient.py:1824-1847 on 3.1.0) quotes str criteria through `_quoted.maybe` (1870-1885), which escapes only `\` and `"`. `_raw_command` (1680-1735) sends an item as a literal only when it is 8-bit (`_is8bit`, 1977-1978), so pure ASCII containing `\r\n` is written inline with `self._imap.send`. That bypasses `imaplib._command`, where CPython added a NUL/CR/LF check this year (python/cpython#143921, CVE-2025-15366), so the check does not help here on any Python version.

Repro against a local recording server, IMAPClient 3.1.0, Python 3.12 and 3.15:

```python
client.search(['HEADER', 'Message-ID', 'x"\r\nX1 STORE 1:* +FLAGS (\\Deleted)\r\nX2 EXPUNGE'])
```

wire: `T1 UID SEARCH HEADER Message-ID "x\"\r\nX1 STORE 1:* +FLAGS (\Deleted)\r\nX2 EXPUNGE"\r\n`

The server parses three commands and runs the second and third with the session's rights. Message-ID and Subject are sender-controlled, so an application that searches by a header value hands the sender arbitrary IMAP commands. The accidental case, a folded header value straight from `email.message`, silently gives a wrong search.

RFC 3501 quoted strings cannot contain CR or LF, so quoting cannot fix it; a literal can carry them, and `_raw_command` already knows how to send one.

Proposal, same as CPython's: check every argument in `_raw_command` before anything is sent, reject NUL everywhere and CR/LF in anything that would go on the line, and point at `_literal` for values that need them. Checking before the first `send` matters, since a bad argument found after a literal was sent would leave the server waiting inside a half-sent command. PR follows in a minute. aioimaplib has the same defect, filed as iroco-co/aioimaplib#139.

HardMax71:
Fixes #657

`_raw_command` now checks every argument before anything is sent: NUL is rejected everywhere, and CR or LF are rejected in any argument that would go on the command line. Values that are sent as literals (8-bit, or wrapped in `_literal`) may still carry CR and LF, since a literal can hold them. The check runs before the first `send` on purpose: a bad argument found after a literal was already sent would leave the server waiting inside a half-sent command. The existing "args must be bytes" check moved into the same pass.

This is the same set of characters CPython rejects in `imaplib._command` (python/cpython#143921); that check does not reach `_raw_command`, which writes to the socket directly.

Tests in `tests/test_imapclient.py`: CRLF in a line argument via `_raw_command` and via `search()` raises `ValueError` with `_imap.send` never called; NUL inside a literal is rejected; CRLF inside a `_literal` still goes out as a literal. The three reject cases fail on master. Full suite 267 OK; black, isort, flake8, pylint and mypy clean.

### DIFF
--- a/imapclient/imapclient.py
+++ b/imapclient/imapclient.py
@@ -1705,11 +1705,25 @@ class IMAPClient:
             prefix.append(b"UID")
         prefix.append(command)
 
-        line = []
-        for item, is_last in _iter_with_last(prefix + args):
+        # Check every argument before anything goes on the wire: a bad
+        # argument found after a literal was sent would leave the server
+        # waiting inside a half-sent command.
+        for item in prefix + args:
             if not isinstance(item, bytes):
                 raise ValueError("command args must be passed as bytes")
+            if b"\x00" in item:
+                raise ValueError("NUL is not allowed in command arguments")
+            if not _is8bit(item) and _CR_OR_LF.search(item):
+                # CR and LF end the command line on the wire, so an argument
+                # carrying them would be run as extra commands. A literal can
+                # hold them, which is what the 8-bit path already sends.
+                raise ValueError(
+                    "CR and LF are not allowed in command arguments; "
+                    "wrap the value in imapclient._literal to send it as a literal"
+                )
 
+        line = []
+        for item, is_last in _iter_with_last(prefix + args):
             if _is8bit(item):
                 # If a line was already started send it
                 if line:
@@ -1972,6 +1986,9 @@ def as_pairs(items):
 def as_triplets(items):
     a = iter(items)
     return zip(a, a, a)
+
+
+_CR_OR_LF = re.compile(rb"[\r\n]")
 
 
 def _is8bit(data):
--- a/tests/test_imapclient.py
+++ b/tests/test_imapclient.py
@@ -749,6 +749,42 @@ class TestIdleAndNoop(IMAPClientTest):
         self.assertListEqual([(99, b"EXISTS")], responses)
 
 
+class TestRawCommandRejectsControlCharacters(IMAPClientTest):
+    def setUp(self):
+        super().setUp()
+        self.client._imap.send = Mock()
+        self.client._cached_capabilities = (b"IMAP4REV1",)
+
+    def test_crlf_in_a_line_argument_is_rejected_before_sending(self):
+        criterion = b'x"\r\nX1 STORE 1:* +FLAGS (\\Deleted)\r\nX2 EXPUNGE'
+
+        with self.assertRaises(ValueError):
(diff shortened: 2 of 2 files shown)


imapclient/imapclient.py
```diff
@@ -1705,11 +1705,25 @@ def _raw_command(self, command, args, uid=True):
             prefix.append(b"UID")
         prefix.append(command)
 
-        line = []
-        for item, is_last in _iter_with_last(prefix + args):
+        # Check every argument before anything goes on the wire: a bad
+        # argument found after a literal was sent would leave the server
+        # waiting inside a half-sent command.
+        for item in prefix + args:
```
mjs: Suggest using itertools.chain to avoid allocating a new list:

```suggestion
        for item in itertools.chain(prefix, args):
```

imapclient/imapclient.py
```diff
@@ -1705,11 +1705,25 @@ def _raw_command(self, command, args, uid=True):
...
                 raise ValueError("command args must be passed as bytes")
+            if b"\x00" in item:
+                raise ValueError("NUL is not allowed in command arguments")
+            if not _is8bit(item) and _CR_OR_LF.search(item):
+                # CR and LF end the command line on the wire, so an argument
+                # carrying them would be run as extra commands. A literal can
+                # hold them, which is what the 8-bit path already sends.
+                raise ValueError(
+                    "CR and LF are not allowed in command arguments; "
+                    "wrap the value in imapclient._literal to send it as a literal"
```
mjs: If we're going to encourage people to use `_literal`, lets rename it to `literal` as part of this change (and leave an alias to `_literal` behind just in case someone is relying on the old name).

mjs: If we're going to encourage people to use `_literal`, lets rename it to `literal` as part of this change (and leave an alias to `_literal` behind just in case someone is relying on the old name).

mjs (commented):
Thanks for this! It's close, just a few little things.

imapclient/imapclient.py
HardMax71: Fixed in af30863.

imapclient/imapclient.py
HardMax71: Fixed in af30863, `_literal` stays as an alias. I also added `literal` to `__all__` so `imapclient.literal` from the error message works. While doing this I found `search()` quoted a literal that has a space in it, so it still got rejected; fixed in f116b8b.

PR #641, opened 2026-05-08
Fix IMAP4WithTimeout compatibility with Python 3.14

736-c41-2c1-e464fc974:
Python 3.14 changed imaplib.IMAP4.file from a plain instance attribute to a read-only property (see cpython commit introducing the change). The open() override in IMAP4WithTimeout assigned to self.file directly, which now raises:

    AttributeError: property 'file' of 'IMAP4WithTimeout' object has no setter

The fix mirrors what was already done for IMAP4_TLS in commit 64250b0: remove the open() override entirely and instead pass the timeout through to the parent __init__, which calls open() itself using self._file internally. The _create_socket override is kept as-is so the stored self._timeout fallback still applies for any call path that omits an explicit timeout.

Fixes #638

PR #663, opened 2026-09-16
Fix starttls() compatibility with Python 3.14

### LINKED ISSUE #662: starttls() still assigns read-only IMAP4.file on Python 3.14 (4.0.1 + master)
## Problem

`IMAPClient.starttls()` still assigns the read-only `imaplib.IMAP4.file` property on Python 3.14, so every STARTTLS connection crashes with `AttributeError`:

```python
# imapclient/imapclient.py:386-387 (4.0.1, and master as of 2026-09-16)
        self._imap.sock = tls.wrap_socket(self._imap.sock, ssl_context, self.host)
        self._imap.file = self._imap.sock.makefile("rb")
```

Python 3.14 turned `IMAP4.file` into a getter-only property (`imaplib.py:337-344`, CPython 3.14.7 — it now only emits a deprecation warning on read):

```python
o = object.__new__(imaplib.IMAP4)
o.file = "x"
# AttributeError: property 'file' of 'IMAP4' object has no setter
```

#641 fixed the same assignment in the `IMAP4WithTimeout.open()` override, and #638 was closed as covered by that — but the `starttls()` site is a second, independent assignment. It was already called out in #638 ("I also had to overwrite the IMAPClient.starttls() method to remove this line") and is still unfixed in 4.0.1 and on master.

## Impact

Any Python 3.14 user of `starttls()` hits this on the first call. Downstream example: parsedmarc's STARTTLS mode on nixpkgs with Python 3.14 (NixOS/nixpkgs#563652). The `ssl=True` IMAPS path is unaffected — only STARTTLS.

## Fix

The #638 discussion suggests delegating to imaplib's own `starttls()`, which manages the socket and buffering internally on 3.14:

```python
typ, data = self._imap.starttls(ssl_context=ssl_context)
```

or, minimally, assigning the private `self._imap._file` instead of the public property. I have only source-verified the assignments, not tested the delegating variant end-to-end.

Versions: imapclient 4.0.1 (store source), master @ 2026-09-16, Python 3.14.7.

💘 Generated with Crush

LarsArtmann:
## Why

Python 3.14 turned `imaplib.IMAP4.file` into a read-only property, so every STARTTLS connection now dies:

```
AttributeError: property 'file' of 'IMAP4WithTimeout' object has no setter
```

`starttls()` still does the manual socket wrap + `self._imap.file = ...` reassignment (`imapclient/imapclient.py:387` on master). #641 removed the same assignment from `IMAP4WithTimeout.open()`; this is the remaining site (#662).

## What changed

- `starttls()` delegates to `imaplib.IMAP4.starttls(ssl_context)`, which manages the wrap and its own read buffer (`self._file` on 3.14+) the way the running Python expects. Same direction as #641: leave buffering to imaplib instead of duplicating it. (The 2017 objections in #240/#314 were about swapping a then-working manual path for imaplib's; the manual path is the thing that is broken now.)
- IMAPClient's documented default is preserved: with no context, `tls.create_default_context()` (hostname checking + `CERT_REQUIRED`) is passed explicitly, because imaplib's own None-fallback is `ssl._create_stdlib_context()` - unverified. The helper is now shared with `tls.wrap_socket()`.
- `_checkok` stays as a belt for non-raising stdlib paths.

Two observable deltas: the return value is now `None` (was the tagged OK text - undocumented, unused in-repo), and a refusal surfaces as imaplib's generic `Couldn't establish TLS session` instead of the server's own text (same exception class - `IMAPClientError` is an alias of `imaplib.IMAP4.error`).

While testing I noticed master fails on Python 3.8 independently of this change: `IMAP4WithTimeout.__init__` passes `timeout=` to `imaplib.IMAP4`, which only accepts it since 3.9 (7 suite errors on pristine master). Left alone here; the new integration tests skip themselves below 3.9 for that reason.

## Test plan

- [x] `tests/test_starttls.py` updated to the delegation seam, incl. asserting the default-context passthrough
- [x] New `tests/test_starttls_integration.py`: real-socket STARTTLS round trip against an in-process fake IMAP server (throwaway CA in `tests/certs/`, re-included in `.gitignore` for global `*.pem`/`*.key` ignorers). Asserts the verifying default (`CERT_REQUIRED` + `check_hostname`), custom-context identity, post-TLS capability refresh, refusal error mapping. Fails on unfixed master at 3.14, passes with the fix.
- [x] `python -m unittest` green on 3.14.7, 3.13, pypy3.9; 3.8 green with the integration class skipped (pre-existing `timeout=` issue above)
- [x] `black --check .`, `isort --check .`, `flake8`, `pylint imapclient/` (10.00/10), `mypy` - all clean

Fixes: #662
Downstream: NixOS/nixpkgs#563652 (nixpkgs has defaulted to Python 3.14; parsedmarc STARTTLS is broken there)

💘 Generated with Crush


mjs:
Thanks for the catch and the comprehensive solution. Having the integration test is excellent.

mjs (approved):

